PR #5, opened 2019-05-29
Fix mistype

jonaselan:


lorisleiva:
Hi, thanks but it’s not a typo. Simply how it is written outside North America.

jonaselan:
Sorry about that. With a simple search I would have realized that. It reminds me that English from North America is not universal. :smile:

PR #16, opened 2019-07-29
Fix Class 'Lorisleiva\Actions\EventDispatcherDecorator' not found

authanram:
Fix Class 'Lorisleiva\Actions\EventDispatcherDecorator' not found on composer install, update and dump-autoload etc. for case sensitive file systems, caused by a typo in the vendor name of the package.


lorisleiva:
Nice one, thanks! 🍺

PR #18, opened 2019-08-31
Laravel 6.0 Support

dennisoderwald:
All tests passed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "illuminate/support": "5.7.*|5.8.*"
+        "illuminate/support": "5.7.*|5.8.*|6.0.*"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


dennisoderwald:
There is a conflict: `Trait method middleware has not been applied, because there are collisions with other trait methods on Lorisleiva\Actions\Action`.

It seems to have something to do with the job middlewares in Laravel 6.0. (https://laravel-news.com/job-middleware-is-coming-to-laravel-6)

lorisleiva:
Thanks for this PR. I had to dig into a few things to resolve the middleware conflicts so it's been committed directly to master.

There are a few more things I'd like to test before creating a new release but you can get it with `dev-master` until then.

Also note that (because of that middleware addition to the Queue class) Laravel actions cannot support <=5.8 and >=6.0 at the same time.

PR #19, opened 2019-09-11
added static init and static exec methods

submtd:
Added static init and exec methods to allow a little cleaner one liner implementation.

**init method**
```
$action = Action::init($attributes);
// instead of 
// $action = new Action($attributes);
```

**exec method**
```
$results = Action::exec($attributes);
// instead of
// $results = (new Action($attributes))->run()
```

### DIFF
--- a/src/Action.php
+++ b/src/Action.php
@@ -5,7 +5,6 @@ namespace Lorisleiva\Actions;
 use BadMethodCallException;
 use Illuminate\Support\Str;
 use Illuminate\Bus\Queueable;
-use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there 👋 

Thanks for this contribution. 

I'm not a big fan of the first one (`init`). I find it confusing to have a static method that simply instantiates a new object.

However, I do see the attraction for the second one (`exec`). I just think the method name is a bit too generic and trying to be a synonym of `run` which can be confusing for new users. What you're doing is running an action as an object.

Currently, you can achieve this with:

```php
$results = (new Action($attributes))->run();
$results = (new Action)->run($attributes);
$results = (new Action)($attributes);

// If the action is not queued then you can also run it as a job.
$results = Action::dispatch($attributes);
```

In any case, I am happy to merge a static method for this purpose but we need to come up with something better than `exec`. Ideally, it would be also `run` but that's impossible in PHP.

mortenscheel:
Perhaps `Action::make($attributes)->run()` could be a compromise?

lorisleiva:
Yeah, that's not a bad idea.

lorisleiva:
Okay I've actually find a way to make `MyAction::run([...])` work as well as `$myAction->run([...])` using `__callStatic()`. Moving this to #36.

PR #33, opened 2020-01-30
Avoid instantiating every listener to check if is a subclass of Action.

phcostabh:
I notice that in every request `laravel-actions` were instantiating every single listener in order to verify if it was a sub-class of `Lorisleiva\Actions\Action`. 

Using PHP's `is_a ` to verify if the listener is an `Action` in the `EventDispatcherDecorator` might reduce the overhead of resolving it from the application container along with all of its dependency in every request.

### DIFF
--- a/src/EventDispatcherDecorator.php
+++ b/src/EventDispatcherDecorator.php
@@ -38,44 +38,44 @@ class EventDispatcherDecorator implements DispatcherContract
             return false;
         }
 
-        return $this->container->make($class) instanceof Action;
+        return false !== in_array(Action::class, \class_parents($class));
     }
-    
+
     public function hasListeners($eventName)
     {
         return $this->dispatcher->hasListeners($eventName);
     }
-    
+
     public function subscribe($subscriber)
     {
         return $this->dispatcher->subscribe($subscriber);
     }
-    
+
     public function until($event, $payload = [])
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there, 👋

Thanks for creating this PR. It's a good catch, I did not know that `is_a` could accept a string as a first argument.

However, I've got a concern with this approach. Reading through the comments on `is_a`, I can see that running it with a string as a first argument will call `__autoload()` on this argument (see [comment](https://www.php.net/manual/en/function.is-a.php#105684)). However, a) I do not know how heavy this `__autoload()` call is in comparison to instantiation and b) this function has been deprecated from PHP 7.2 (see [doc](https://www.php.net/manual/en/function.autoload.php)).

Do you have any thought on this?

phcostabh:
Although the [is_a](https://github.com/php/php-src/blob/8f4f1dea34e255eff8ff4d13960d287d809665a6/Zend/zend_builtin_functions.c#L795) function calls the autoloader only if the class is not already loaded, calling the `__autoload` itself is bad enough.

But it might be safe to use `intsanceof` instead. 

Let me know what you think.

lorisleiva:
I do like that `class_parents` method! Simple and elegant. Well done 👍

PR #34, opened 2020-02-28
Laravel 7 support

exodusanto:
Update `illuminate/support` version

-----
Waiting for Laravel 7 release


lorisleiva:
Hi there 👋

Thanks for that but I've got to do some more thorough check before upgrading to Laravel 7.

A new release should be tagged today :)

PR #51, opened 2020-08-03
Add `static` docblock for run method

imliam:
Per the docs for [running an action as an object](https://laravelactions.com/actions-as-objects.html), the `run` method can be called either on the instance or as a static method, so the docblocks should represent both so that tooling and static analysis can pick up on it.

### DIFF
--- a/src/Action.php
+++ b/src/Action.php
@@ -14,6 +14,7 @@ use ReflectionMethod;
 
 /**
  * @method mixed run(...$mixed)
+ * @method static mixed run(...$mixed)
  */
 abstract class Action
 {


lorisleiva:
Hi Liam 👋 

I originally wanted to do this too but PHPStorm was complaining about the duplicate name for the same method. I also noticed that having only the non-static one still autocompleted after `::` which is why I left it like this.

But, to be honest, I'm not sure if that's the right way to do this so I'm happy to take your advice on that. 🙂

Let me know if you still think it's the right thing to do and I'll merge this! Thank you 🙏

innocenzi:
I don't know about PHPStorm, but VSC is complaining because of the missing docblock:

imliam:
Yeah - it was VSCode Intelephense complaining that prompted me to mention this, wasn't aware that PHPStorm would reject it. Not really sure of a middleground for them both.

imliam:
FYI, this issue on Jetbrains' tracker: https://youtrack.jetbrains.com/issue/WI-13583

lorisleiva:
Let's keep this PR open until PHPStorm fix this issue and then we'll get that merged. 💪

PR #59, opened 2020-09-10
added laravel 8 support

Jam0r85:


lorisleiva:
I'm afraid this is not enough to upgrade to Laravel 8. Testbench also needs to be updated for the tests to work. I also need to make sure tests are showing up in PRs because for some reason they do not anymore. I'll try to get all that sorted this weekend.

lorisleiva:
Fixed by #60

PR #64, opened 2020-10-10
Add missing import to RunsAsCommand

szepeviktor:
and remove a useless `@var`

### DIFF
--- a/src/Concerns/RunsAsCommand.php
+++ b/src/Concerns/RunsAsCommand.php
@@ -2,6 +2,7 @@
 
 namespace Lorisleiva\Actions\Concerns;
 
+use Exception;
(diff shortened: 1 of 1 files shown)


szepeviktor:
trait RunsAsCommand -> static function registerCommand -> Closure -> `$command = $this;`

@lorisleiva This is complex and not testable with PHPStan, so this code relies on us humans, and that is error prone 😿 
Please consider rewriting it. 🚀 Straight-forward code is beneficial.

lorisleiva:
Hi there, unfortunately the most non-intrusive way to register an action as a command for now is to wrap it in a closure.

Regarding this PR, the Exception import is actually uses and without it tests are failing. 😕

PR #66, opened 2020-10-16
Fixes ignored default values in handle method.

### LINKED ISSUE #61: Attributes dynamically resolved from handle method do not use default value
Consider the following action.

```php
class MyAction extends Action
{
    protected $getAttributesFromConstructor = true;

    public function handle($myAttributeWithDefaultValue = true)
    {
        return $myAttributeWithDefaultValue;
    }
}
```

Current behaviour.

```php
MyAction::run();
// => null
```

Expected behaviour.

```php
MyAction::run();
// => true
```

Root cause: default value is hardcoded to `null` here.

https://github.com/lorisleiva/laravel-actions/blob/ef190a20c2d270d1907e2df612a20a2eb8d19dfa/src/Action.php#L77-L79

jstoks:
Fixes #61 

- Always call `resolveAttributesFromConstructor` in the constructor to account for possible
  attributes in handle signature.
- Treat attributes in `handle` and attributes in `getAttributesFromConstructor` separately.
- Normalize return value to `$this` in `resolveAttributesFromConstructor`

### DIFF
--- a/src/Action.php
+++ b/src/Action.php
@@ -54,9 +54,8 @@ abstract class Action
             $this->resolveAndCall($this, 'initialized');
         }
 
-        if (func_num_args() > 0) {
-            $this->resolveAttributesFromConstructor(...func_get_args());
-        }
+        $args = func_num_args() > 0 ? func_get_args() : [];
+        $this->resolveAttributesFromConstructor(...$args);
     }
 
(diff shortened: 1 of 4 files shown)


lorisleiva:
Nice! Thank you so much for that! 🍺

Issue #54, opened 2020-08-25
Ignore `null` for command output


slashequip:
When using an action as a command, the default behaviour, without the `consoleOutput` method, is to output the return value of the `handle` method.

In a lot of cases, at least in my uses, there is no return value and then the command just outputs `null` without formatting causing line inconsistencies, it'd be great if the output was ignored when `null`, as it's a nothing value anyway, and nothing output to the console.

jstoks:
I have this little PR #67. I thought it would be a quick one, but it's not obvious how to test this.

With the tools available in Laravel, I'm not sure how to test this case. There is no `expectsNoOutput`. I looked at the framework source and I just don't see a good option for implementing it.

Any ideas?

lorisleiva:
Fixed by #67 and released as `v1.1.7`. 🚀

PR #67, opened 2020-10-21
Ignore console output when null

jstoks:
Trying to take care of #54.

- When there is no output from a command do not write any output.
- Having trouble testing this case.


lorisleiva:
Hi there 👋 

Thanks for that!

Regarding the tests, since: `var_export(null)` outputs the string `'NULL'`, I think it's just the case of making sure the output is an empty string.

Perhaps `->expectsOutput('');` should do the trick.

jstoks:
Thanks for the suggestion.

I'm not very familiar with PHP I should have looked `var_export` up myself. My laziness is showing.

lorisleiva:
Perfect! Thank you for your contribution! 🍺

PR #68, opened 2020-11-19
Ensure models are serialized correctly within attributes

### LINKED ISSUE #55: Job serialization for models
When queuing an action as a job the actions data is not serialized like we would see by default in a Laravel job, I believe this is due to `attributes` being an array and Laravel's `SerializesModels` trait only looks at the top level property and not at the values of an array.

To achieve this I believe we could loop through the attributes array during the Laravel Actions `SerializesModels` `__sleep` method and apply the same logic Laravel does for each of those properties, same applies for the`__wakeup` method to, for rehydrating this data.

slashequip:
Closes #55.

This PR adds additional prep work during `sleep` and `wakeup` methods to correctly serialize models and collections within the attributes array.

These changes will ensure the payload of jobs remain as small as possible.

These changes _may_ be breaking in some cases. Previously a model would have been serialized like a regular class, meaning should you have changed some model attributes and not persisted them then the job would be run with those un-persisted changes. After these changes, the model will be fetched from the database, fresh, when the model is unserialized. It's unlikely but a consideration non-the-less.

### DIFF
--- a/src/Concerns/SerializesModels.php
+++ b/src/Concerns/SerializesModels.php
@@ -9,6 +9,8 @@ trait SerializesModels
     use BaseSerializesModels {
         __sleep as protected sleepFromBaseSerializesModels;
         __wakeup as protected wakeupFromBaseSerializesModels;
+        __serialize as protected serializeFromBaseSerializesModels;
+        __unserialize as protected unserializeFromBaseSerializesModels;
     }
 
     public function __sleep()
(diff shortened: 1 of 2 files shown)


slashequip:
@lorisleiva these test failures explain why Laravel has both `__sleep` and `__serialize` in the trait. Seemingly PHP 7.3 prefers `__sleep` whilst 7.4 prefers `__serialize`. Will dig into it, should be a similar process but the methods work slightly differently.

lorisleiva:
😩 Ergh. Sorry to hear that.

Thank you for looking into it. ❤️ 

Tests are 👌

slashequip:
@lorisleiva looks like the addition of `__serialize` and `__unserialize` methods were needed for 7.4 we are green if you want to give this a final review.

lorisleiva:
Thank you! 🍺

Issue #41, opened 2020-05-21
Use stubs to generate Actions through the make command


lorisleiva:
Just opening an issue to make sure I don't forget to use Laravel 7's new stub feature for generated Actions.

PR #78, opened 2021-01-07
Use stubs to generate Actions through the make command

ahmedabdel3al:
This PR  allow us to stubs to generate actions through the make command  #41 

**`php artisan vendor:publish --provider="Lorisleiva\Actions\ActionServiceProvider" --tag=stubs`**


lorisleiva:
Thanks for that! Just a heads up: Laravel Actions v2 (being released very soon) has been re-written from scratch and no longer offers the `make:action` command since there's no longer a forced structure for your action.

But that PR can still be valuable for users who might want to stick to v1. 🙂

PR #80, opened 2021-01-08
Laravel Actions v2

lorisleiva:


cyrillkalita:
I did read [Upgrade guide](https://laravelactions.com/2.x/upgrade.html) and I think I understand where you are coming from.

Granted, I was not using any of the fancy stuff; running as Jobs and controllers, as I really to have a class that does just one thing: _a single-action class_.

I really liked that I can keep on adding data to the action class as I was going through the motions elsewhere - and I could validate it once I am done with it, to run the action - which was saving a ton of time, as final validation and handling was now was always done at the last mile of the journey.

With version two a new paradigm that  is introduced makes _my_ flow harder, rather then easier; I am sure it is not so for the rest of the people out there. I just venting. 

Would you consider supporting 1.x beyond L8?

lorisleiva:
Hi there 👋 

Thanks for the detailed feedback. I don't think the plan would be to continue to actively maintain both version. Of course if the upgrade from v1 to L9 is just a composer update — which will likely be the case — then that's absolutely fine.

However, I'd prefer if we could improve on Laravel Actions v2 to remove these pain points whilst keeping its current "lack of opinion".

From what I read it seems like your main pain point is with regard to the fact that there is no longer a set of unified attributes across pattern and so authorization and validation is reserved to controllers.

I've written a reply to someone else who shares the same frustration that you can find here.

https://github.com/lorisleiva/laravel-actions/issues/97

Basically, it's still very much on the back of my mind and I want to find a way to provide this again for v2 as an optional trait.

Aside from this, is there anything else that you would miss from Laravel Actions v1?

cyrillkalita:
Every since I got introduced I to single action classes you're package been my lodestar, I wouod be honest. Every project I've done, relies on it. So yeah, you summed it up pretty well.

Thank you, your contribution is valuable and appreciated.

lorisleiva:
Hi @cyrillkalita 👋 

Just to let you know, version `2.1` comes with a new optional trait `WithAttributes` that brings back the ability to authorize and validate actions using attributes for all patterns — i.e. as objects, as controllers, as jobs, etc.

I've also added [an entire page dedicated to that trait in the documentation](https://laravelactions.com/2.x/use-unified-attributes.html). I hope this helps.

PR #90, opened 2021-02-11
Ignore view template files when registering action routes

onemoreahmad:
Hi,

I got an error when I put my views files within the same folder with actions when registering the routes.

Screen Shot 2021-02-11 at 3 14 43 AM

Screen Shot 2021-02-11 at 3 15 52 AM

This is a suggested fix to check if the file is not a view.

### DIFF
--- a/src/ActionManager.php
+++ b/src/ActionManager.php
@@ -132,9 +132,13 @@ class ActionManager
         }
 
         foreach ((new Finder)->in($paths->toArray())->files() as $file) {
-            $this->registerRoutesForAction(
-                $this->getClassnameFromPathname($file->getPathname())
-            );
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there 👋 

Sorry for the long reply. Thanks for raising this.

I think it'd be better if we could have a more generic check when going through files.

Inside the `registerRoutesForAction` method, we could add a check (before the 3 current ones) that uses reflection to check if the class exists.

lorisleiva:
Hi Ahmad 👋 

I had some free time so I decided to tackle it and released as part of the next release which is  [`v2.1.1`](https://github.com/lorisleiva/laravel-actions/releases/tag/v2.1.1). 🚀 

Let me know if that all works fine for you. Thanks again for raising this! 🍺

Issue #92, opened 2021-02-23
registerRoutes not working with forward slash on windows


FrankVisuals:
When calling 
```
Actions::registerRoutes('app/Actions/Test');
```
i have the issue

```
ErrorException: class_parents(): Class App\C:\<my-path>\app\Actions\Test\Login does not exist and could not be loaded 
```

The issue here is that the path was not parsed correctly for me (using Windows). It only works when calling 

```
Actions::registerRoutes('app\\Actions\\Test');
```

lorisleiva:
Hi there 👋 

Thanks for raising this and having a go at it.

If I remember correctly, I got heavily inspired with how Laravel auto-registers the Commands directory for this feature.

Does this work well for your on Windows? If so I might have missed something there. I'd really appreciate your help in this since I have no Windows computer to test that. 😕 

Thanks again. 🙂

FrankVisuals:
Hey! Thanks for the fast response. I tried to provide a PR with a fix, however it seems to fail on Linux (tests) - it worked for me (https://github.com/lorisleiva/laravel-actions/pull/93). I can try to give you as many details as possible.

I narrowed it down to the `ActionManager` class where `Str::after($pathname, realpath(app_path()) . DIRECTORY_SEPARATOR)` results in `"C:\<path to my project>\app/Actions/Test\CustomLogin.php"` when being called like `Actions::registerRoutes('app/Actions/Test');`

The actual result here should be `Actions\Test\CustomLogin.php` - which works when I use `Actions::registerRoutes('app\\Actions\\Test');`

Hope that helps!

lorisleiva:
Hey, yeah I saw but PR #93 will only work on Windows. The same goes when using backslashes in your path — i.e. `Actions::registerRoutes('app\\Actions\\Test');`. Even in your case you'll probably end up deploying this code on a linux server which will fail at this point.

I've double-checked [the `load` method on the Laravel Console Kernel](https://github.com/laravel/framework/blob/b69db31d47342f41878c20cbdce35edc8ea56147/src/Illuminate/Foundation/Console/Kernel.php#L208-L236) and I'm using the exact same process to discover files in an array of directories.

[...]

PR #93, opened 2021-02-23
fix issue with path

FrankVisuals:
This is meant to resolve the issue described in #92 

The slashes in the pathname had to be replaced first in order to fix the behaviour for windows.


FrankVisuals:
Sadly the tests were working on my windows mashine - seems like it wasn't a valid fix :(

FrankVisuals:
Awesome - thanks for the fix!

Issue #116, opened 2021-05-25
Allow to publish stubs


Wulfheart:
When running `php artisan stub:publish` the action stub doesn't get exported.

Wulfheart:
I will open a PR with a fox later that day.

Wulfheart:
I found out that `php artisan stub:publish` only publishes Laravel's stubs. `php artisan vendor:publish --tag=stubs` does the trick.

PR #119, opened 2021-05-30
Refactoring stub variables to match Laravel's pattern

Wulfheart:
Compare https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Console/stubs/test.stub. The generation still works, I didn't find a test for it.
Related to #116.

### DIFF
--- a/src/Console/stubs/action.stub
+++ b/src/Console/stubs/action.stub
@@ -1,10 +1,10 @@
 <?php
 
-namespace DummyNamespace;
+namespace {{ namespace }};
 
 use Lorisleiva\Actions\Concerns\AsAction;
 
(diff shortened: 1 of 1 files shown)


lorisleiva:
I do love consistency. Thanks!

Wulfheart:
Maybe a new minor release (as it is only a bugfix) would be good for this change to arrive at the end user who just runs `composer require lorisleiva/laravel-actions`.

lorisleiva:
Technically, both naming conventions are recognised by the framework so the current behaviour is not broken but I do like to make sure we use the same convention as the framework should they decide to deprecate the old one. 🙂

Issue #110, opened 2021-05-03
Form Display and Update


keyurshah:
Hello,

This package looks great! Thank you for your hard work.

In many of my use cases, I have a form that shows a button to execute an action, i.e. SyncInventory

So, it seems I would need 2 actions, one to display the form, and one to execute the action. 

Instead of creating 2 separate actions, is there a simpler way to create this pattern inside one single action?

- show form - show a view that has the base form, perhaps a dedicated function that can be used as a GET, `asForm()` ? 
- execute form on POST - call `asController()` - (current functionality)

Thank You

lorisleiva:
Hi there 👋 

It’s a good point actually. I usually use two actions: one for displaying the form and one for handling it because I like my actions to be that granular.

However, I think there’s a way to do it in one action. I haven’t tested this but if you give a different method when defining the route it should use that one instead. For example:

```php
Route::get('MyAction@asForm');
```

Let me know if that works for you, otherwise I’m sure we can work something out. 😊

keyurshah:
Appreciate your help!

I tried to do it 2 different ways, but it goes straight to the `handle()` method. 

```
Route::get('/test1', '\App\Actions\UploadInventory@asForm');
Route::get('/test2', [\App\Actions\UploadInventory::class,'asForm']);
```

my action looks like this

```

   public function asForm() {
        dd('inside asForm');
   }

   public function handle() {
       dd('inside handle')
   }
```

separately, in your current what naming pattern would you currently use for the form showing this button? Since it's not a true action, it seems challenging to name it as an action.

lorisleiva:
Oh okay thanks for letting me know. I’ll have a look at how I can update the ControllerDecorator to use that information.

It usually depends on the form but I tend to prefix them with “Show” so it’s quite explicit. For example, I might have “UpdateUserProfileAction” and “UpdateUserPasswordAction” as form handlers, and then “ShowUserSettingsAction” as the action that displays both forms in the same page.

lorisleiva:
Hi there 👋 

Sorry for the delay but, from [`v2.1.3`](https://github.com/lorisleiva/laravel-actions/releases/tag/v2.1.3), you should now be able to provide explicit methods to your route and they will be called within your action (see #123). 🙂

keyurshah:
Sorry I missed this earlier. It looks perfect and I'll try this out. Thank you so much for putting in the time for this request!

PR #123, opened 2021-06-17
Support controllers with explicit methods

lorisleiva:
See #110


Wulfheart:
Is this feature documented yet?

lorisleiva:
No it's a good point but I'm not sure this is the sort of behaviour we should encourage for Laravel Actions since you end up with multiple actions within the same class. But it is definitely something that should work if they try it so that's why I made the fix. Do you think we should document it still?

Wulfheart:
But especially for showing a form and handling that it would be really awesome.

lorisleiva:
Alrighty, I'll add something on the guide page. 🙂

Wulfheart:
But we could also discuss if there is or should be a better way to handle this use case?

lorisleiva:
I've written this: https://laravelactions.com/2.x/register-as-controller.html#routes-with-explicit-methods

And I've just made sure my note at the bottom is correct haha.

Do you mean you'd like to mention alternative ways to handle the show/process forms duality? Like using `ShowX` actions as I mentioned on issue #110?

Wulfheart:
Oh yes. This one is gold.

lorisleiva:
Documentation updated 👌

Issue #161, opened 2022-01-25
Handle a job failure


abrampl:
Is it  possible to handle failed job in laravel actions like in orginal laravel queue, method failed ?

https://laravel.com/docs/8.x/queues#dealing-with-failed-jobs

```php

    /**
     * Handle a job failure.
     *
     * @param  \Throwable  $exception
     * @return void
     */
    public function failed(Throwable $exception)
    {
        // Send user notification of failure, etc...
    }
```

travisaustin:
Within my own projects, I defined a new `JobDecorator` that I use to do this. In my `JobDecorator`, it looks for a function called `jobFailed()` on the Action itself, and calls that function if the job does fail.

I've submitted a PR to add this functionality to Laravel Actions. I hope this helps!

lorisleiva:
Fixed by #166. Will be part of the next release (very soon, feel free to use `dev-main` for now 🙂).

alex-osborn:
Hi. This is a great update - thanks.

I'm running into an error with the current release version (v2.4.0), while running my project which is using PHP 8.0.18. This is a valid version per the package requirements (^8.0).

    Cannot unpack array with string keys

Looks like this is the line causing the error:

    Lorisleiva\Actions\Decorators\JobDecorator:150

    $this->fromActionMethod('jobFailed', [$e, ...$this->parameters], []);


Looks like this pull request introduced a dependency on PHP 8.1, because array unpacking with associative keys was only introduced in that version. https://php.watch/versions/8.1/spread-operator-string-array-keys

Possibly bump the version for the next release or look at tweaking this?

lorisleiva:
Hi there 👋 Thanks for raising this. I think this is probably more a change in Laravel since I've never seen this `parameters` array be associative. Would you mind opening another issue for this?

PR #166, opened 2022-01-28
Support for new function jobFailed(Throwable $e)

travisaustin:
When a Laravel job fails, Laravel supports cleaning up from that job by defining a `failed()` function on the job class.

See Laravel documentation here: https://laravel.com/docs/8.x/queues#cleaning-up-after-failed-jobs

This PR adds support for a new function on the Laravel Action called `jobFailed(Throwable $e)`. If a job fails, this function will be called to provide an opportunity to clean up from the job failure.

See #161.

I don't have much experience writing tests, and would appreciate any feedback on writing a better test if needed.


lorisleiva:
Awesome! Thanks for your contribution! 🍺

gwillyoo:
This is not found in the documentation or am i missed it  ? But thanks!

Issue #173, opened 2022-02-26
Conditional runs


Wulfheart:
I just stumbled across this and thought it would be a good idea to make running an action as an object analogous to running it as a job.

Therefore I propose to add `runIf(bool, ...)` and `runUnless(bool, ...)` (similar to `dispatch`) to the AsObject syntax so something like this isn't necessary anymore:
```php
if($someCondition){
    Action::run();
}

// would become this
Action::runIf($someCondition);
```

lorisleiva:
I like this! 👌

Do you think we can get away with a minor semver update for it? Technically it is a breaking change as we're adding methods to a trait but I doubt this will have an real impact on devs. Wdyt?

Wulfheart:
I don't think that it is a major breaking change.
As per [semver.org](https://semver.org/):

> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards compatible manner, and
> 3. PATCH version when you make backwards compatible bug fixes.

Therefore I think it is only a minor version as it only adds functionality in a backward compatible manner. I am going to open a PR for it.

lorisleiva:
Thanks! I meant more like, if someone has already defined the `runIf` or `runUnless` methods and the signatures don't match then it will break their actions but the likelihood of that is very low so I'm happy with a minor jump. 🙂

lorisleiva:
Released in `v2.4.0`! 🔥

PR #174, opened 2022-02-27
Conditional Object runs added

Wulfheart:
This refers to #173 and adds the methods `SomeAction::runIf($boolean)` and `SomeAction::runUnless($boolean)`. There are also some tests. The implementation is analogous to the `AsJob` trait.

As these changes are backward compatible I think it warrants only a minor version bump.


lorisleiva:
That's perfect, thanks! I'll spend a bit of time trying to close some issues today and I'll tag a new version afterwards. 🍺

PR #189, opened 2022-08-06
change abstract class to interface 

salarmotevalli:
in this situation i think it's better to use interface instead of abstract class because only visibility and return type are identified


lorisleiva:
Hey there 👋 

Thanks for that but I'm not looking to make this refactoring at this time. I will certainly consider it for the next version upgrade though. 🙏

PR #194, opened 2022-09-18
Added ajax response method.

georgemjohn:
Added ajax response method


leandrodiogenes:
Im sorry but what is the difference of an ajax response and json response?

Apparently laravel treats it the same way.

```php
namespace Illuminate\Http\Concerns;

use Illuminate\Support\Str;

trait InteractsWithContentTypes
{
    ...
    public function expectsJson()
    {
        return ($this->ajax() && ! $this->pjax() && $this->acceptsAnyContentType()) || $this->wantsJson();
    }
   ...
}
```

leandrodiogenes:
Please take a test.
Simulate an action with both methods in the same class.
jsonResponse and ajaxResponse.
I believe one will conflict with the other because both are treated the same by laravel

PR #215, opened 2022-12-03
Update AsController.php

ddunford:
Using LaravelActions with dingo API causes an exception, as it calls getMiddleware(), which expects an array when using array_merge.

### DIFF
--- a/src/Concerns/AsController.php
+++ b/src/Concerns/AsController.php
@@ -21,6 +21,6 @@ trait AsController
      */
     public function getMiddleware()
     {
-        // ...
+        return [];
     }
 }


Wulfheart:
What is dingo API?

ddunford:
> What is dingo API?

https://github.com/api-ecosystem-for-laravel/dingo-api 

It allows you to build API's. It calls getMiddleware() and expects an array Laravel controller returns an array. 

Your examples show getMiddleware() returns an array anyway, so it makes sense to have the default return an empty array to stop any type issues, like this one.

lorisleiva:
Hi there 👋 

I think that makes sense.

Could you please update the docblocks accordingly as it says it returns `void` at the moment?

ddunford:
Updated :)

lorisleiva (approved):

lorisleiva:
Thank you!

PR #217, opened 2023-01-06
AsListener EventDiscovery

GameSpecU:
I often use EventDiscovery in my projects - I missed it in this package :)
Actually, native EventDiscovery could discover Actions, but it registers events based on `handle` method parameters instead of `asListener`

The feature works pretty much the same as Laravel's Event Discovery - it uses **Lody** instead of **Finder** as a difference

- I used an additional `EventServiceProvider` instead of the main `ActionServiceProvider` because laravel uses only these services in the `event:cache` command
- Since we are implementing all cases in the Action class, those without Event Class in `handle` method parameter are also registered

- Discovering Events may looks weird, because half is done by Collection, but second using hard-coded array flip.
I tried different ways to make it done using Collection grouping but decided to leave it as it is because of performance reasons & complex/confusing operations - Maybe somebody knows a better way :)

Other changes:
- Updated Orchestra TestBench, because the previous version throws an error in `package:discovery` command
- The command above was needed to register dependent ServiceProviders (Lody in this particular case)

TODO
- [ ] Readme
- [ ] Refactor & Publish EventServiceProvider for per-project customization such as path-resolving
- [ ] Consider doing the above instead of using the config file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,8 @@
         "lorisleiva/lody": "^0.3.0"
     },
     "require-dev": {
-        "orchestra/testbench": "7.0.0",
-        "orchestra/testbench-core": "7.0.0",
+        "orchestra/testbench": "^7.0.0",
+        "orchestra/testbench-core": "^7.0.0",
         "pestphp/pest": "^1.2",
         "phpunit/phpunit": "^9.5"
     },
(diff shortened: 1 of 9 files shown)


Wulfheart:
Tbh I am not sure if a second service provider is the correct way forward here.

Issue #202, opened 2022-10-18
Actions don't work with Octane


jedjones-uk:
Hi,

I've been testing actions for use with octane and it seems they don't work properly. The first run of an action goes to the html output, and the second and all subsequent requests to the same endpoint goes to the json output.

I think it has to do with the fact that `ControllerDecorator@__invoke` is not called subsequent requests but I've not really been able to trace it down to a specific set of code.

The subsequent requests also seem to go direct to the action itself.

Spice-King:
Going to tag in with the same problem.

As far as I can see, `ActionManager`'s `$extended` property is not getting cleared between runs, thus on the second run is not calling `app()->extend()`. Going to see if I find a solution and see about making a PR.

Spice-King:
Looks like I was right. Debating on how to make a PR, but this is the hack job of a listener does work when added to `config/octane.php` for the `RequestReceived` event if you're interested @jedjones-uk.
```php
<?php

namespace App\Listeners;

use Laravel\Octane\Events\RequestReceived;
use Lorisleiva\Actions\ActionManager;

class FixLaravelActionsOnOctane
{
    public function handle(RequestReceived $event)
    {
        $manager = $event->app->get(ActionManager::class);
        $r = new \ReflectionObject($manager);
        $p = $r->getProperty("extended");
        $p->setAccessible(true);
        $p->setValue($manager, []);
    }
}

```
Octane really seems to lack documentation about best practices for making packages with it, outside of this [bit on dependacy injection](https://laravel.com/docs/9.x/octane#dependency-injection-and-octane). Just no mention of the events or the included listeners for them that reset the pins for Laravel and it's first party packages.

jedjones-uk:
@Spice-King 

This is great work for tracing this down to a repeatable solution, I spent a while losing hair trying to figure out just where the key issue was. You fix works fine as a hack but a long term quality solution I think requires 2 changes.

1: Define some helper functions to determine if octane is in use like these: https://github.com/laravel/octane/issues/508#issuecomment-1096937174

2: when octane is in use, skip the `$this->isExtending($abstract)` call in the `extending` function for `ActionManager`

This would avoid the need for reflection and keeps everything in house.

Thanks again.

ClaraLeigh:
Been having weird routing issues lately. Dependancy and route binding working 99% of the time but occasionally failing. Turning off octane seems to have cleaned it up, wonder if its related to this

Wulfheart:
[...]

PR #221, opened 2023-02-16
bugfix(routing): Resolve a routing issue found in octane

ClaraLeigh:
## Description

When using octane, I found that subsequent requests would often fail, with two different errors showing:

> App\xxxxx::handle(): Argument #1 ($model) must be of type App\Model\Xxxxx, string given, called in /var/task/vendor/lorisleiva/laravel-actions/src/Concerns/AsController.php on line 14

> Too few arguments to function App\xxxxx::handle(), 0 passed

From these errors, I can determine that route binding and dependancy injection is breaking on occasion. As such we need to ensure that we are resolving the arguments correctly before calling the handle() method.

## Changes Made

File `src/Concerns/AsController.php`
- Add a new execute() fn that essentially runs the ControllerDecorator again. This is not ideal but works for now.

File `src/Decorators/ControllerDecorator.php`
- Ensure that we have substituted route bindings, _this seems to be the main fix for the problem_
- Added an `execute` method to call the default run function

## Issues remaining

I feel like there is a double up now, that ControllerDecorator runs twice as it seems to resolve things like jsonResponse and htmlResponse. 
This is why the new `execute` function uses `run` and not `callAction` or `__invoke`

## Testing - needs work

This needs to be tested in a replicable octane environment, for the moment I tested by manually disabling the call to `replaceRouteMethod` in the decorator's constructor and then updating the `AsController::__invoke` method to look like so:

```
$decorator = (new ControllerDecorator($this, \Route::current()));
$decorator->replaceRouteMethod();
return $decorator->execute();
```

This would then cause the default `AsController::__invoke` function to run first, first like it does in subsequent Octane calls

## Related Issues

- https://github.com/lorisleiva/laravel-actions/issues/202

### DIFF
--- a/src/Concerns/AsController.php
+++ b/src/Concerns/AsController.php
@@ -2,6 +2,9 @@
 
 namespace Lorisleiva\Actions\Concerns;
 
+use Lorisleiva\Actions\Decorators\ControllerDecorator;
+use Illuminate\Support\Facades\Route;
+
 trait AsController
 {
     /**
@@ -11,7 +14,7 @@ trait AsController
      */
     public function __invoke(...$arguments)
     {
-        return $this->handle(...$arguments);
+        return (new ControllerDecorator($this, Route::current()))->execute();
     }
 
     /**
--- a/src/Decorators/ControllerDecorator.php
+++ b/src/Decorators/ControllerDecorator.php
@@ -54,6 +54,13 @@ class ControllerDecorator
         }, $this->middleware);
     }
 
+    public function execute()
+    {
+        return $this->run(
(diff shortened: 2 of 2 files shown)


lorisleiva:
> essentially runs the ControllerDecorator again. This is not ideal but works for now.

Hi there, thank you for tackling this issue but could we solve this differently? I don't think running something twice can be an option.

ClaraLeigh:
@lorisleiva yea I would like to but the reason for putting everything here is the hope that someone might work out what the real problem is. A bit low on time to dig into it further at this time

PR #223, opened 2023-02-18
Add jsonApiResponse method in the controller

digitlimit:
This PR will add jsonApiResponse which will handle requests going to /api endpoints 
I noticed there is no way to do this if I want to use the same action for both web and api


lorisleiva:
Hi there, sorry I have no plans to support this. Thank you for understanding.

PR #225, opened 2023-02-21
Add option to configure custom job decorator classes

### LINKED ISSUE #224: Allow overriding the job decorator classes
This is related to #145

I'd like to create a PR that allows users to override which decorator classes should be used when using actions as jobs. This would allow a package like Venture, for example, to provide first-party integration with Laravel Actions without the user having to change their actions in any way.

The API could be similar to what Cashier–and for that matter, [Venture itself](https://laravel-venture.com/configuration/customizing-models.html)–does to allow users to provide custom models. Here's a suggestion of what this might look like:

```php
// in some service provider

ActionManager::useJobDecorator(MyCustomJobDecorator::class);
ActionManager::useUniqueJobDecorator(MyCustomUniqueJobDecorator::class);
```

Venture could then provide a `Venture::useLaravelActions()` method to register the necessary decorators in order to make the two packages compatible with each other. Users should then be able to use actions from this package as jobs in their workflows without any other changes.

The only thing that really needs to change as far as Laravel Actions is concerned, is replacing the hardcoded references to `JobDecorator` and `UniqueJobDecorator` with something like `ActionManager::$jobDecoratorClass`.

Let me know if this works for you and I can submit a PR for this!

ksassnowski:
## Summary

This PR adds the ability to define custom job decorator classes. These custom classes will then get used when using an action as a job. The default implementation is to use the existing `JobDecorator` and `UniqueJobDecorator` classes, so this should be a backwards compatible change.

## Code changes

In order to change the decorator classes, two new methods were added to the `ActionManager`.

- `ActionManager::useJobDecorator(string $jobDecoratorClass)` changes the decorator class that gets used when an action gets turned into a non-unique job
- `ActionManager::useUniqueJobDecorator(string $uniqueJobDecoratorClass)` changes the decorator that gets used when an action gets turned into a unique job

Apart from that, the only non-test code that needed be changed were the `makeJob`, `makeUniqueJob`, and `assertPushed` methods in the `AsJob` trait. These methods will now grab the decorator class from the `ActionManager` at runtime.

## Tests

Since `$jobDecoratorClass` and `$uniqueJobDecoratorClass` are both static properties on the `ActionManager`, I decided to play it extra safe and add a global `afterEach` callback to the test suite which resets them back to their default values. This avoids having to remember to manually reset this in every test case that might change these default settings (as has happened to me when implementing this).

I also added two datasets to more easily check that a given test passes both for the default decorator classes, as well as custom decorator classes.

## Additional Info

I have already tested these changes in a large real-world application and can confirm that they are sufficient to ship first-party integration with Laravel Actions inside Venture. Let me know what you think 😊 

Closes #224

### DIFF
--- a/src/ActionManager.php
+++ b/src/ActionManager.php
@@ -3,15 +3,24 @@
 namespace Lorisleiva\Actions;
 
 use Illuminate\Console\Application as Artisan;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
 use Illuminate\Routing\Router;
 use Lorisleiva\Actions\Concerns\AsCommand;
 use Lorisleiva\Actions\Concerns\AsController;
 use Lorisleiva\Actions\Concerns\AsFake;
+use Lorisleiva\Actions\Decorators\JobDecorator;
+use Lorisleiva\Actions\Decorators\UniqueJobDecorator;
 use Lorisleiva\Actions\DesignPatterns\DesignPattern;
 use Lorisleiva\Lody\Lody;
 
 class ActionManager
 {
+    /** @var class-string<JobDecorator> */
+    public static string $jobDecorator = JobDecorator::class;
+
+    /** @var class-string<JobDecorator&ShouldBeUnique> */
+    public static string $uniqueJobDecorator = UniqueJobDecorator::class;
+
     /** @var DesignPattern[] */
     protected array $designPatterns = [];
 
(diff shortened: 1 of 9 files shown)


Wulfheart:
I can’t say anything to the technicalities yet. But the communication style is 💅

ksassnowski:
I'm glad it's appreciated. I spend a lot of time on these PR descriptions 😅

ksassnowski:
The test failures seem to be unrelated to my changes as I get the same failures on `main`.

src/ActionManager.php
```diff
@@ -23,6 +32,22 @@ public function __construct(array $designPatterns = [])
...
+    /**
+     * @param class-string<JobDecorator> $jobDecoratorClass
+     */
+    public static function useJobDecorator(string $jobDecoratorClass): void
+    {
+        static::$jobDecorator = $jobDecoratorClass;
+    }
+
+    /**
+     * @param class-string<JobDecorator&ShouldBeUnique> $uniqueJobDecoratorClass
```
ksassnowski, line 44: These are the type constraints that Laravel Actions actually expects this class to have. I double checked this with a local repository to see that PHPStan would actually be able to check for this correctly.

lorisleiva (approved):
That's great work, thanks @ksassnowski! 🍻 

Side note for a different PR: do you think it would make sense to support custom decorators for other patterns like controllers, listeners, etc?

PR #233, opened 2023-04-06
feat: increase backtrace frame limit

jaulz:
Currently, the limit of backtrace frames is set to 10 and cannot be changed. It would be great if it would be possible to increase the limit for custom design patterns (e.g., I would like to use it in combination with Lighthouse). Anyway, while debugging the same I was wondering whether the first 9 frames could be skipped because those all belong to `laravel` or `laravel-actions` and thus the time could rather be spent on real potential matches.

### DIFF
--- a/src/ActionManager.php
+++ b/src/ActionManager.php
@@ -28,6 +28,8 @@ class ActionManager
     /** @var bool[] */
     protected array $extended = [];
 
+    protected int $backtraceLimit = 10;
+
     public function __construct(array $designPatterns = [])
     {
         $this->setDesignPatterns($designPatterns);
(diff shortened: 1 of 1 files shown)


jaulz:
@lorisleiva sorry to disturb but would you be able to give some feedback on this? 😊

lorisleiva:
Hi there, sorry for the late reply. I'm a bit worried about skipping backtrace frame we don't control in case something changed in the framework but I'd be happy to skip the ones that are purely from the package as we control that part. Would you be able to show me the frames we'd be skipping?

Also you mention in your OP that you'd like the limit to be parameterised. I can see the limit as an argument in this method but perhaps you'd like this to be accessible at a higher level somewhere? Is this what you meant?

jaulz:
@lorisleiva thanks for the quick response! 😊 So this is the stacktrace:
image

Essentially there are only 2 frames from your library so it would only be a minor performance gain. However, I just pushed an update that reduces the frames anyway and also adds a method to change the limit. The limit can be adapted like this:
```php
    $this->app->extend(ActionManager::class, function (ActionManager $actionManager) {
      $actionManager->setBacktraceLimit(20);

      return $actionManager;
    });
```

What do you think?

jaulz:
For some reason the commit does not appear here but it's actually pushed to https://github.com/jaulz/laravel-actions/tree/patch-1 🤔

lorisleiva (approved):
Thanks!

PR #247, opened 2023-07-23
Add support for listener shouldQueue method

dmason30:
This  PR adds support for the below functionality that exists on traditional listeners.

> #### Conditionally Queueing Listeners
> 
> Sometimes, you may need to determine whether a listener should be queued based on some data that are only available at runtime. To accomplish this, a `shouldQueue` method may be added to a listener to determine whether the listener should be queued. If the `shouldQueue` method returns `false`, the listener will not be executed:
> ```php
>     <?php
> 
>     namespace App\Listeners;
> 
>     use App\Events\OrderCreated;
>     use Illuminate\Contracts\Queue\ShouldQueue;
> 
>     class RewardGiftCard implements ShouldQueue
>     {
>         /**
>          * Reward a gift card to the customer.
>          */
>         public function handle(OrderCreated $event): void
>         {
>             // ...
>         }
> 
>         /**
>          * Determine whether the listener should be queued.
>          */
>         public function shouldQueue(OrderCreated $event): bool
>         {
>             return $event->order->subtotal >= 5000;
>         }
>     }
> ```
> _Source: https://laravel.com/docs/10.x/events#conditionally-queueing-listeners_

### DIFF
--- a/src/Decorators/ListenerDecorator.php
+++ b/src/Decorators/ListenerDecorator.php
@@ -26,6 +26,15 @@ class ListenerDecorator
         }
     }
 
+    public function shouldQueue(...$arguments)
+    {
+        if ($this->hasMethod('shouldQueue')) {
+            return $this->resolveFromArgumentsAndCall('shouldQueue', $arguments);
+        }
+
+        return true;
+    }
+
     protected function resolveFromArgumentsAndCall($method, $arguments)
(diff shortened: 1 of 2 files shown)


lorisleiva (approved):
Thanks for the contribution and for providing a test. 🍺

PR #248, opened 2023-07-27
feat: allow custom design patterns

jaulz:
This PR allows custom patterns to be registered via:
```php
    $this->app->extend(ActionManager::class, function (ActionManager $actionManager) {
      $actionManager->registerDesignPattern(new FooBarDesignPattern());

      return $actionManager;
    });
```

### DIFF
--- a/src/ActionManager.php
+++ b/src/ActionManager.php
@@ -61,6 +61,13 @@ class ActionManager
         return $this->designPatterns;
     }
 
+    public function registerDesignPattern(DesignPattern $designPattern): ActionManager
+    {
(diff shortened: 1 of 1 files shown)


lorisleiva:
Thanks! FYI, you can also use the Facade to update the Action Manager like so.

```php
use Lorisleiva\Actions\Facades\Actions;

Actions::registerDesignPattern(new FooBarDesignPattern());
```

lorisleiva (approved):

jaulz:
Ah, thanks for the merges and the hint 😊

PR #250, opened 2023-08-11
feat: add static runWithAttributes method

jaulz:
This helper method allows users to run an action with attributes and validation.

### DIFF
--- a/src/Concerns/WithAttributes.php
+++ b/src/Concerns/WithAttributes.php
@@ -5,6 +5,8 @@ namespace Lorisleiva\Actions\Concerns;
 use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
 use Lorisleiva\Actions\AttributeValidator;
+use ReflectionMethod;
+use ReflectionParameter;
 
 trait WithAttributes
 {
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there, I'm not too sure about this approach to be honest because it assumes the Action's attributes and the handle's arguments are the same which they may not be.

jaulz:
Ah, sorry, I missed to pass the required arguments. Now, only the arguments that are found on the method's signature will be passed to the handle method.

nickfls:
@jaulz as far as I can tell, you can achieve something like this with:

```php
ActionClass::make()
    ->fill(['foo' => 'bar'])
    ->set('variable', value)
    ->handle(['other' => 'attributes'])
````

I sometimes use it when I want syntax sugar and better readability, like `AddUserProfile` around `StoreUserProfile`:
```
/**
     * @param \App\Models\Contracts\Profile|\App\Models\Profile $profile
     * @param array $attributes
     * @return \App\Models\Contracts\UserProfile
     */
    public function handle(Profile $profile, $attributes = []): UserProfile
    {
        return StoreUserProfile::make()
            ->fill($attributes)
            ->handle(['profile_id' => $profile->getKey()]);
    }
```
but i recognize that your use case may be different

PR #264, opened 2023-11-26
replace laravel application to lumen

elsayed85:


Wulfheart:
And Lumen is more or less discontinued. Why use it then?

PR #273, opened 2024-03-11
Support Laravel 11

patrickomeara:
Only support two latest major versions of Laravel as per Support Policy https://laravel.com/docs/11.x/releases#support-policy

Remove Laravel 9 and PHP 8.0 support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,13 +22,13 @@
     ],
     "require": {
         "php": "^8.1",
-        "illuminate/contracts": "^10.0 || ^11.0",
-        "lorisleiva/lody": "dev-l11-compatibility"
+        "illuminate/contracts": "^10.0|^11.0",
+        "lorisleiva/lody": "^0.5"
     },
     "require-dev": {
-        "orchestra/testbench": "^8.5",
-        "pestphp/pest": "^1.23 || ^2.0",
-        "phpunit/phpunit": "^9.6 || ^10.0 || ^11.0"
+        "orchestra/testbench": "^9.0",
+        "pestphp/pest": "^1.23|^2.34",
+        "phpunit/phpunit": "^9.6|^10.0"
     },
     "repositories": [
         {
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,8 @@
         }
     ],
     "require": {
-        "php": "^8.0",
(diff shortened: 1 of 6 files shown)


jameswagoner:
This PR looks proper compared to what Laravel Shift attempted

https://github.com/laravel-shift/laravel-actions/blob/l11-compatibility/composer.json#L25

nickfls:
@patrickomeara would you consider adjusting `composer.json` with

```json
 "phpunit/phpunit": "^9.6 || ^10.0"
````

**Comment by patrickomeara** [comment]
Good idea @nickfls I have allowed latest versions of Pest and PHPUnit. I leave it up to @lorisleiva to decide whether to release a new major version and remove the lower constraints.

File: composer.json [hunk 1]
```diff
@@ -21,15 +21,21 @@
         }
     ],
     "require": {
-        "php": "^8.0",
-        "illuminate/contracts": "9.0 - 9.34 || ^9.36 || ^10.0",
-        "lorisleiva/lody": "^0.4"
+        "php": "^8.1",
+        "illuminate/contracts": "^10.0 || ^11.0",
+        "lorisleiva/lody": "dev-l11-compatibility"
```
lorisleiva, line 26: Lody needs to be upgraded first.

lorisleiva (changes requested):
Hi there, thanks for this PR but I'm afraid it's not mergeable because its dependencies need to upgrade first.

The reason Lody is not upgraded yet is because Testbench is still under the `-dev` tag which makes the Shift PR fail.

I can resolve that manually but I'm afraid I've got very little time for these packages right now.

composer.json
patrickomeara: Hi @lorisleiva 

I noted that in the commit message

```
Use Compatible Lody version

This commit can be removed and the correct version of lody use after https://github.com/lorisleiva/lody/pull/6 has been merged
```

I'm happy for you to use this branch or close it. Whatever is easier.

Thanks for your work.

oliverearl:
Apologies that it's off-topic, but just wanted to drop in and sincerely thank everyone for getting this through so quickly! We use this library extensively both at work and on the side and this helps us get an early start on Laravel 11 upgrades. 🥳

patrickomeara:
Thanks for your work on this @lorisleiva

PR #283, opened 2024-08-07
Handle $deleteWhenMissingModels on JobDecorator

patrickomeara:
This enables the user to configure the job using `->setDeleteWhenMissingModels()` within `configureJob()` or by setting `$jobDeleteWhenMissingModels = true`


lorisleiva (approved):
Thanks! Looks good to me. I'm not familiar with this option but I'm assuming this is available on both Laravel 10 and 11 otherwise CI should fail.

patrickomeara:
Hey Loris! Yeah it's in L10. https://laravel.com/docs/10.x/queues#ignoring-missing-models

Thanks for your hard work with this package.

PR #284, opened 2024-08-09
Fix Orchestra Laravel 10 compatibility

edwinvdpol:
When you run PHP 8.1, and you try `composer install` the following error will appear:

Screenshot 2024-08-09 at 22 54 32
Laravel 10 [needs](https://packages.tools/testbench#version-compatibility) Orchestra Testbench 8.

This PR fixes that compatibility.

The branch name is not entirely correct, but you get the point. 😋

👋 Edwin

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
         "controller",
         "job",
         "object",
-        "command"
+        "command",
+        "listener"
     ],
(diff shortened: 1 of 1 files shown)


lorisleiva (approved):
Yeah technically I only run the L10 tests on CI which has a matrix override but you're right this is more correct. :)

PR #286, opened 2024-08-13
Add assert with params convenience methods 

patrickomeara:
### Problem

When I am testing that actions are pushed correctly, I almost always want to check the parameters that it is pushed with.

Currently it is a little cumbersome to check the params, as they are passed in as an array.

```php
SendPodcastPublishedEmail::dispatch($user, $podcast);

// currently...
SendPodcastPublishedEmail::assertPushed(fn ($a, $params) => $params[0]->is($user) && $params[1]->is($podcast));
```
Additionally, the thing I want to check is in the second argument so I have an unused variable.

### Solution
If we spread the params into the callback we can use type safety to check we have the correct type and shortcut the param check.

```php
SendPodcastPublishedEmail::dispatch($user, $podcast);

// assert by closure
SendPodcastPublishedEmail::assertPushedWithParams(fn (User $u, Podcast $p) => $user->is($u) && $podcast->is($p));
```

Or we can assert by an array
```php
// assert by array
SendPodcastPublishedEmail::assertPushedWithParams([$user, $podcast]);
```

Or my favourite way if the action only has a single param.
```php
SendWelcomeEmail::dispatch($user);

// assert by first class callable
SendWelcomeEmail::assertPushedWithParams($user->is(...));
```

This PR also adds the ability to assert the queue using `assertPushedWithParamsOn()`, and assert that an action wasn't pushed with params using `assertNotPushedWithParams()`

I later found a test helper `assertJobPushedWith()` that could be replaced.

Note: Naming the methods `assertPushedWith()`, `assertNotPushedWith()` also works well, but sounds a little strange when checking the queue as well `assertPushedWithOn()` I'm easy either way.

### DIFF
--- a/src/Concerns/AsJob.php
+++ b/src/Concerns/AsJob.php
@@ -169,4 +169,29 @@ trait AsJob
             return $callback ? $callback(...func_get_args()) : true;
         });
     }
+
+    public static function assertPushedWithParams(Closure|array $callback, ?string $queue = null): void
+    {
+        if (is_array($callback)) {
+            $callback = fn (...$params) => $params === $callback;
+        }
+
+        static::assertPushed(
+            fn ($action, $params, JobDecorator $job, $q) => $callback(...$params) && (is_null($queue) || $q === $queue)
+        );
+    }
+
+    public static function assertPushedWithParamsOn(string $queue, Closure|array $callback): void
+    {
+        static::assertPushedWithParams($callback, $queue);
+    }
+
+    public static function assertNotPushedWithParams(Closure|array $callback): void
+    {
+        if (is_array($callback)) {
+            $callback = fn (...$params) => $params === $callback;
+        }
+
+        static::assertNotPushed(fn ($action, $params, JobDecorator $job) => $callback(...$job->getParameters()));
+    }
 }
--- a/src/Concerns/AsJob.php
+++ b/src/Concerns/AsJob.php
@@ -170,7 +170,7 @@ trait AsJob
         });
     }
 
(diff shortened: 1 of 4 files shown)


lorisleiva:
Hiya! 👋 

Thanks for the great PR!

I'm really liking this and I love the fact that this kills a test helper. It's usually a good sign that this is needed for the API. However, I'm having a hard time digesting the name `assertPushedWithParamsOn` haha.

Also, I'm not sure if this is now the convention in Laravel, but in testing frameworks, I'm more used to a single `With` suffix instead of `WithParams`. I think the following API could be a bit more intuitive.

- Rename `assertPushedWithParams` to `assertPushedWith`. E.g. `SendPodcastPublishedEmail::assertPushedWith([$user, $podcast]);` looks very clean to me.
- Same with `assertNotPushedWith`.
- Kill `assertPushedWithParamsOn`. We already have `assertPushedWith` and `assertPushedOn` so why not just make people do two assertions in order to compose them instead of trying to create helpers that do it all. See example below.

```ts
SendPodcastPublishedEmail::assertPushedWith([$user, $podcast]);
SendPodcastPublishedEmail::assertPushedOn("someQueue");
```

What do you think?

patrickomeara:
I think you're right about the naming, I'll update that and clean it up. Let me have a think about the queue, because having it in two different assertions could assert two different jobs have been pushed.

patrickomeara:
The queue can be asserted using 

```php
SendPodcastPublishedEmail::assertPushedWith([$user, $podcast], 'someQueue');
```

lorisleiva:
Perfect, thank you for this! 🍺

PR #285, opened 2024-08-09
Cleanup docblocks added return types

edwinvdpol:
Hi there,

Don't be shocked by the amount of files changed, but I felt like doing the following:

* Remove obsolete docblocks, and corrected a few.
* Added type hints where possible.
* Replaced `switch` with `match` (just looks better / cleaner imho)
* Added the new `$jobDeleteWhenMissingModels` property / getter to `AsJob.php`

There were quite a few return types missing in the test files, that's why it looks so much.

Hope you like it, and feel free to contact me if you have any questions / remarks.
You are free to close this PR if you don't like it of course, but static analysis tools will be (more) happy eventually. 😉

👋 Edwin

PR #287, opened 2024-08-19
Remove return type for make()

Casmo:
Remove static return type. See https://github.com/lorisleiva/laravel-actions/pull/285#issuecomment-2296463178

### DIFF
--- a/src/Concerns/AsObject.php
+++ b/src/Concerns/AsObject.php
@@ -6,7 +6,7 @@ use Illuminate\Support\Fluent;
 
 trait AsObject
 {
-    public static function make(): static
+    public static function make()
     {
         return app(static::class);
     }
--- a/tests/AsActionTest.php
+++ b/tests/AsActionTest.php
@@ -85,6 +85,12 @@ it('runs as an object', function () {
     expect(AsActionTest::$latestResult)->toBe(42);
 });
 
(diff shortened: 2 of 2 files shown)


Casmo:
Done!

lorisleiva:
Thanks!

austincarpenter:
This has broken IDE autocompletion for me. Now PhpStorm thinks `handle()` when called via `::make()->handle(...)` points to `vendor/laravel/framework/src/Illuminate/Foundation/Application.php`'s `handle()` method.

I deliberately don't use the `::run()` syntax for this reason.

lorisleiva:
@austincarpenter Hi there, this was a partial revert of https://github.com/lorisleiva/laravel-actions/pull/285 which was only published a `patch` version earlier (in the matter of days). When did you experience that autocompletion working and then not working?

austincarpenter:
@lorisleiva Prior to #285 it was working due to the docblock, also after due to the return type and now post #287 it's not due to neither a docblock nor a return type.

lorisleiva:
Gotcha, so we just need the docblock back right? If so, would you be able to submit a small PR for this and I'll get that merged/patched immediately?

austincarpenter:
Apologies for the delay on this one. Thanks @TaylorWilton @lorisleiva 👏

PR #292, opened 2024-09-09
Add a Return Type using PHPDoc to fix IDE Autocomplete

TaylorWilton:
Hi 👋 

I've also been affected by the IDE autocompletion issue noted in #285 and #287.

image
image

I figured since it was a simple PR and no-one else had done it yet I may as well create it.

My understanding of the issue is that the make function can't define a return type, because the result could be a Decorator, or the actual action. However, without docblocks or type-hinting, PHPStorm ends up providing the autocompletion for the `handle` function defined in the application container, which is different. 

I've returned the docblock here, which resolves the issue for me locally, and won't cause any type errors. 

Thanks for your work maintaining this package!


lorisleiva (approved):
Tysm! 🫶

PR #302, opened 2024-12-20
Add shouldExpect() mocking method

### LINKED ISSUE #278: Idea: add shouldExpect() to an action mock
@lorisleiva there is a `static::shouldRun()` method, which is a wrapper around `static::mock()->shouldRecieve('handle')`

Would you have any reservations to add `shouldExpect()`.. something like:

```php
    /**
     * @return Expectation|ExpectationInterface|ExpectsHigherOrderMessage
     */
    public static function shouldExpect()
    {
        return static::mock()->expects('handle');
    }
```

I like how `expects` allows you to both set expectations and evaluate it after..?

LucasBrunner:
This PR adds the static `shouldExpect` method to the `AsFake` trait.

This exposes the behavior of Mockery's [`->expects()`](https://docs.mockery.io/en/latest/reference/alternative_should_receive_syntax.html?highlight=expects#expects). Users will now be able to use `::shouldExpect()` to declare that a mocked action should run once and only once.

Fixes #278

### DIFF
--- a/src/Concerns/AsFake.php
+++ b/src/Concerns/AsFake.php
@@ -41,6 +41,11 @@ trait AsFake
         return static::mock()->shouldReceive('handle');
     }
 
+    public static function shouldExpect(): Expectation|ExpectationInterface|HigherOrderMessage
+    {
+        return static::mock()->expects('handle');
+    }
+
     public static function shouldNotRun(): Expectation|ExpectationInterface|HigherOrderMessage
     {
         return static::mock()->shouldNotReceive('handle');
--- a/tests/AsFakeTest.php
+++ b/tests/AsFakeTest.php
@@ -74,6 +74,19 @@ it('can mock an action expecting for it to run', function () {
     expect($result)->toBe(3);
 });
 
+it('can mock an action expecting for it to run only once', function () {
+    // Given we mock the action with the
+    // expection it should run only once.
+    AsFakeTest::shouldExpect();
+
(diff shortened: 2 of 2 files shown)


lorisleiva:
Thanks for this! I'm just not a great fan of the method name `shouldExpect` because that is not how the fluent API of `expects` is supposed to start with.

The current helper `shouldRun` makes sense because the underlying fluent API starts with `shouldReceive`.

I think `expects` would be better but would be more likely to cause a conflict with existing method names.

In any case this introduces a breaking change by adding a function to everyone's action class and I'm not sure this little change warrants a bump to v3.

@Wulfheart what so you think?

Wulfheart:
@lorisleiva no opinion on that. However, if you say it is inconsistent that I would be against it in its current form and be open for another naming. It still seems useful on first thought.

LucasBrunner:
I agree that the name `shouldExpect` is not great and changing it is no problem. 

It's a great point that since `AsFake` is a trait adding methods to it could result in namespace collisions with the class using it. Since this means a version bump I would like to take the opportunity to suggest a few further additions to improve the testing experience, but those ideas are more substantial changes than this so I will make dedicated issues for them.

lorisleiva:
Sorry guys, I'm gonna make the decision to close this for now as this is a significant breaking change and I don't think the benefits warrant a v2 bump. Let's revisit if more people start to crave this.

PR #304, opened 2025-01-10
Add support for AsPipeline concern

### LINKED ISSUE #279: Feature Request: Support for Illuminate Pipelines
Relevant Article: https://martinjoo.dev/laravel-pipelines

Illuminate pipelines look like:
```php
app(Pipeline::class)
    ->send('<p>This is the HTML content of a blog post</p>')
    ->through([
        ModerateContent::class,
        RemoveScriptTags::class,
        MinifyHtml::class,
    ])
    ->then(function (string $content) {
        return Post::create([
            'content' => $content,
            ...
        ]);
    });
```

In the example above, the `string $content` variable "travels" the pipeline. The signature of each class that it travels through needs to be 
```php
public class ModerateContent
{
    public function handle(/* our traveler */ string $content, Closure $next)
    {
        $safeContent = $this->removeBadWords($content);
        $next($safeContent);
    }
}
```

As you can see, the syntax is fairly similar to the existing Laravel Actions handler. In fact, all I need to do is change the signature to `?Closure $next = null` to adapt it for both usages. However, it would be quite neat if Actions could support this directly (the return value of the handler function would be passed to `$next` in this case).

stevenmaguire:
As discussed in [Feature Request: Support for Illuminate Pipelines #279](https://github.com/lorisleiva/laravel-actions/issues/279) there seems to be some interest and appetite in allowing our Action classes to function well within a Pipeline workflow, in addition to all the other already support Laravel conventions. 

This PR aims to do that. 

The main thing to note are the two opinions being asserted by the new `AsPipeline` trait. An explanation and justification for those opinions are captured in a code comment in the trait file.

If there are any critiques or comments, please share them. I am happy to collaborate to get this to a production-ready and mergable state.

Fixes #279

### DIFF
--- a/src/Decorators/PipelineDecorator.php
+++ b/src/Decorators/PipelineDecorator.php
@@ -22,6 +22,6 @@ class PipelineDecorator
 
         $method = $this->hasMethod('asPipeline') ? 'asPipeline' : 'handle';
 
-        return $closure($this->callMethod($method, [$passable]) ?? $passable);
+        return $closure($this->callMethod($method, [$passable]) ?? $passable) ?? $passable;
     }
 }
--- a/tests/AsPipelineTest.php
+++ b/tests/AsPipelineTest.php
@@ -4,26 +4,25 @@ namespace Lorisleiva\Actions\Tests;
 
 use Illuminate\Support\Facades\Pipeline;
 use Lorisleiva\Actions\Concerns\AsAction;
+use Lorisleiva\Actions\Tests\Stubs\PipelinePassable;
 
 class AsPipelineTest
 {
     use AsAction;
 
-    public function handle(AsPipelinePassable $passable): void
+    public function handle(PipelinePassable $passable): void
     {
         $passable->increment();
     }
 
-    public function asPipeline(AsPipelinePassable $passable): AsPipelinePassable
+    public function asPipeline(PipelinePassable $passable): void
     {
         $this->handle($passable);
-
-        return $passable;
     }
 }
 
 it('can run as a pipe in a pipeline, with an explicit asPipeline method', function () {
-    $passable = Pipeline::send(new AsPipelinePassable)
+    $passable = Pipeline::send(new PipelinePassable)
         ->through([
             AsPipelineTest::class,
             AsPipelineTest::class,
@@ -32,21 +31,57 @@ it('can run as a pipe in a pipeline, with an explicit asPipeline method', functi
         ])
         ->thenReturn();
 
-    expect(is_a($passable, AsPipelinePassable::class))->toBe(true);
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
     expect($passable->count)->toBe(4);
 });
 
 it('can run with an arbitrary via method configured on Pipeline', function () {
-    $passable = Pipeline::send(new AsPipelinePassable)
+    $passable = Pipeline::send(new PipelinePassable)
         ->via('arbitraryMethodThatDoesNotExistOnTheAction')
         ->through([
             AsPipelineTest::class,
-            AsPipelineTest::class,
-            AsPipelineTest::class,
-            AsPipelineTest::class,
+            app()->make(AsPipelineTest::class),
         ])
         ->thenReturn();
 
-    expect(is_a($passable, AsPipelinePassable::class))->toBe(true);
-    expect($passable->count)->toBe(4);
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(2);
+});
+
+it('can run as a pipe in a pipeline with only one explicit container resolved instance at the bottom of the stack', function () {
+    $passable = Pipeline::send(new PipelinePassable)
+        ->through([
+            AsPipelineTest::class, // implicit container resolved instance
+            app()->make(AsPipelineTest::class), // explicit container resolved instance
+        ])
+        ->thenReturn();
+
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(2);
+});
+
+it('cannot run as a pipe in a pipeline with an explicit container resolved instance in the middle of the stack', function () {
+    $passable = Pipeline::send(new PipelinePassable)
+        ->through([
+            AsPipelineTest::class, // implicit container resolved instance
+            app()->make(AsPipelineTest::class), // explicit container resolved instance
+            AsPipelineTest::class, // implicit container resolved instance
+            AsPipelineTest::class, // implicit container resolved instance
+        ])
+        ->thenReturn();
+
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(2);
+});
+
+it('cannot run as a pipe in a pipeline as an standalone instance', function () {
+    $passable = Pipeline::send(new PipelinePassable)
+        ->through([
+            new AsPipelineTest, // standalone instance
+            AsPipelineTest::class, // implicit container resolved instance
+            app()->make(AsPipelineTest::class), // explicit container resolved instance
+        ])
+        ->thenReturn();
+
+    expect(is_null($passable))->toBe(true);
 });
--- a/src/DesignPatterns/PipelineDesignPattern.php
+++ b/src/DesignPatterns/PipelineDesignPattern.php
@@ -17,7 +17,8 @@ class PipelineDesignPattern extends DesignPattern
 
     public function recognizeFrame(BacktraceFrame $frame): bool
     {
-        return $frame->matches(Pipeline::class, 'Illuminate\Pipeline\{closure}');
+        return $frame->matches(Pipeline::class, 'Illuminate\Pipeline\{closure}')
+            || $frame->matches(Pipeline::class, '{closure:{closure:Illuminate\Pipeline\Pipeline::carry():184}:185}');
     }
 
     public function decorate($instance, BacktraceFrame $frame)
--- a/tests/AsPipelineWithExplicitTraitTest.php
+++ b/tests/AsPipelineWithExplicitTraitTest.php
@@ -4,27 +4,26 @@ namespace Lorisleiva\Actions\Tests;
 
 use Illuminate\Support\Facades\Pipeline;
 use Lorisleiva\Actions\Concerns\AsPipeline;
+use Lorisleiva\Actions\Tests\Stubs\PipelinePassable;
 
 class AsPipelineWithExplicitTraitTest
 {
     use AsPipeline;
 
-    public function handle(AsPipelinePassable $passable): void
+    public function handle(PipelinePassable $passable): void
     {
         $passable->increment();
     }
 }
 
 it('can run as a pipe in a pipeline, with explicit trait, without asPipeline method', function () {
-    $passable = Pipeline::send(new AsPipelinePassable)
+    $passable = Pipeline::send(new PipelinePassable)
         ->through([
             AsPipelineWithExplicitTraitTest::class,
-            AsPipelineWithExplicitTraitTest::class,
-            AsPipelineWithExplicitTraitTest::class,
-            AsPipelineWithExplicitTraitTest::class,
+            app()->make(AsPipelineWithExplicitTraitTest::class),
         ])
         ->thenReturn();
 
-    expect(is_a($passable, AsPipelinePassable::class))->toBe(true);
-    expect($passable->count)->toBe(4);
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(2);
 });
--- a/tests/AsPipelineWithImplicitTraitTest.php
+++ b/tests/AsPipelineWithImplicitTraitTest.php
@@ -4,27 +4,26 @@ namespace Lorisleiva\Actions\Tests;
 
 use Illuminate\Support\Facades\Pipeline;
 use Lorisleiva\Actions\Concerns\AsAction;
+use Lorisleiva\Actions\Tests\Stubs\PipelinePassable;
 
 class AsPipelineWithImplicitTraitTest
 {
     use AsAction;
 
-    public function handle(AsPipelinePassable $passable): void
+    public function handle(PipelinePassable $passable): void
     {
         $passable->increment();
     }
 }
 
 it('can run as a pipe in a pipeline, with implicit trait, without asPipeline method', function () {
-    $passable = Pipeline::send(new AsPipelinePassable)
+    $passable = Pipeline::send(new PipelinePassable)
         ->through([
             AsPipelineWithImplicitTraitTest::class,
-            AsPipelineWithImplicitTraitTest::class,
-            AsPipelineWithImplicitTraitTest::class,
-            AsPipelineWithImplicitTraitTest::class,
+            app()->make(AsPipelineWithImplicitTraitTest::class),
         ])
         ->thenReturn();
 
-    expect(is_a($passable, AsPipelinePassable::class))->toBe(true);
-    expect($passable->count)->toBe(4);
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(2);
 });
--- a/tests/AsPipelineWithMultipleNonOptionalParametersTest.php
+++ b/tests/AsPipelineWithMultipleNonOptionalParametersTest.php
@@ -5,17 +5,19 @@ namespace Lorisleiva\Actions\Tests;
 use ArgumentCountError;
 use Illuminate\Support\Facades\Pipeline;
 use Lorisleiva\Actions\Concerns\AsAction;
+use Lorisleiva\Actions\Tests\Stubs\PipelinePassable;
+use TypeError;
 
 class AsPipelineWithMultipleNonOptionalParametersTest
 {
     use AsAction;
 
-    public function handle(AsPipelinePassable $passable, int $nonOptionalAdditionalParameter): void
+    public function handle(PipelinePassable $passable, int $nonOptionalAdditionalParameter): void
     {
         $passable->increment();
     }
 
-    public function asPipeline(AsPipelinePassable $passable): AsPipelinePassable
+    public function asPipeline(PipelinePassable $passable): PipelinePassable
     {
         $this->handle($passable);
 
@@ -24,9 +26,19 @@ class AsPipelineWithMultipleNonOptionalParametersTest
 }
 
 it('cannot run as a pipe in a pipeline expecting multiple non-optional parameters', function () {
-    $passable = Pipeline::send(new AsPipelinePassable)
+    $passable = Pipeline::send(new PipelinePassable)
         ->through([
             AsPipelineWithMultipleNonOptionalParametersTest::class,
+            app()->make(AsPipelineWithMultipleNonOptionalParametersTest::class),
         ])
         ->thenReturn();
 })->throws(ArgumentCountError::class);
+
+it('cannot run as a pipe in a pipeline as an explicit container resolved instance preceding an implicit container resolved instance', function () {
+    $passable = Pipeline::send(new PipelinePassable)
+        ->through([
+            app()->make(AsPipelineWithMultipleNonOptionalParametersTest::class),
+            AsPipelineWithMultipleNonOptionalParametersTest::class,
+        ])
+        ->thenReturn();
+})->throws(TypeError::class);
--- a/tests/Stubs/PipelinePassable.php
+++ b/tests/Stubs/PipelinePassable.php
@@ -1,11 +1,11 @@
 <?php
 
-namespace Lorisleiva\Actions\Tests;
+namespace Lorisleiva\Actions\Tests\Stubs;
 
 /**
  * Test fixture used in all of the AsPipeline{*} tests.
  */
-class AsPipelinePassable
+class PipelinePassable
 {
     public function __construct(public int $count = 0)
     {


src/Concerns/AsPipeline.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function asPipeline(mixed ...$arguments): mixed
+    {
+        $passable = array_shift($arguments);
+        $closure = array_pop($arguments);
+
+        $returned = $this->handle($passable);
+
+        if (! is_null($returned)) {
+            return $closure($returned);
```
edalzell, line 37: the `else` isn't needed here cuz of the `return`

stevenmaguire: True. Thanks for catching that.

stevenmaguire:
> How come the handle method isn't used? The "standard" pattern in this package is to check for the as... and use that if found but if not, use the handle method.

Do you mean in the `PipelineDecorator::__invoke` method? If not, where were you expecting to see that? I'd like to know what context you are asking about before responding to the "why" of your question.

edalzell:
CleanShot 2025-01-10 at 10 08 04@2x

src/Concerns/AsPipeline.php
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return $closure($passable);
```
edalzell, line 40: Another nitpick, but this is personal preference for sure, but this whole `if` could be `return $closure($this->handle($passable) ?? $passable);`

stevenmaguire:
Thanks. That's what I thought you were asking about. Yes, I clocked that implementation detail but originally did not include it because if that method was not available, the Pipeline resolution chain would be broken anyway. I'm not sure how someone would configure things in such a way where the Decorator was in play without also having the trait, which furnishes a `asPipeline` method. But, as I acknowledged I don't know the complete inner workings of the rest of the package and I will take it your might know something I don't, so better safe than sorry. The PR has been updated with more guard rails in place as you appeared to be expecting.

src/Concerns/AsPipeline.php
stevenmaguire: Yes, that could still work and reduce the code footprint.

edalzell:
> I will take it your might know something I don't, so better safe than sorry. 

Nope, but in general you should match how a package works. Consistency is important.

stevenmaguire:
> Nope, but in general you should match how a package works. Consistency is important.

I tend to agree with that, but not as a dogmatic approach. If the code path can't be accessed or exercised with a test (that replicates what a consuming project _could_ do), it shouldn't be included just because another class in the project does it. I'm still scratching my head on how I could exercise that code path, even though I've now added it.

Do you have a suggestion on how that might be exercised with a test?

edalzell:
> I tend to agree with that, but not as a dogmatic approach. 

Totally agree, mine was only a suggestion to be considered. If it isn't right then we shouldn't do it.

> Do you have a suggestion on how that might be exercised with a test?

Not off the top of my head, but this was only a quick review. If it's dead code path then it def shouldn't be included, but I know as a use of this package I would expect having only a `handle` method to work.

I'll take a look later when I have a few moments.

Thanks again for your work on this, it's awesome.

stevenmaguire:
I just tried to test some use cases that would exercise that code path and was unable to get there. 

I tried a pipe class which explicitly used another trait from the package but not `AsPipeline` - I tried `AsController` - and it failed. Likely because the `PipelineDesignPattern` didn't find a match and did not put the decorator in play. 

I tried a generic pipe class and naturally, it did not get there.

In the meantime, I've removed the fallback code path from the decorator.

I would love a review from @lorisleiva or anyone else with close working knowledge of this package. At the moment, there are two main test cases that are passing.

lorisleiva:
Hey guys, thanks for this! I'll have a proper look at it over the weekend and read all your comments.

On little thing I've noticed on a first quick look is that the `asPipeline` method is defined on the trait. Other patterns tend to make the decorator resolve from the `asX` method first and fallback to the `handle` method. I can see the `asPipeline` as some specific requirements but it would be nice for the same mental model to apply here.

Sorry if thats mentioned in your conversation though, as I said I'll have proper read through everything soon.

stevenmaguire:
Thanks @lorisleiva!

Regarding the `asPipeline` method being on the trait, I explained the justification in the code comment on the method itself. 

```php
    /**
     * Typical pipeline behavior expects two things:
     *
     *     1)  The pipe class to expect a single incoming parameter (along with
     *         a closure) and single return value.
     *     2)  The pipe class to be aware of the next closure and determine what
     *         should be passed into the next pipe.
     *
     * Because of these expectations, this behavior is asserting two opinions:
     *
     *     1)  Regardless of the number of parameters provided to the asPipeline
     *         method implemented here, only the first will be supplied to the
     *         invoked Action.
     *     2)  If the invoked Action does not return anything, then the next
     *         closure will be supplied the same parameter. However, if the
     *         invoked action does return a non-null value, that value will
     *         be supplied to the next closure.
     *
     * Also, this logic is implemented in the trait rather than the decorator
     * to afford some flexibility to consuming projects, should the wish to
     * implement their own logic in their Action classes directly.
     */
```

Basically, the whole value of the concern here is that you (the consuming package) don't need to worry about handling the Pipeline's expectations around the callback closure, this concern will take care of it for you. 

So, the opinionated logic that is now here in the PR should be furnished by the package somewhere. If that is the case, then furnishing it in the trait leaves it accessible to consuming projects to override the default behavior if they please. 

Without this opinionated logic in the package it will be left to the consuming project to implement every time and if they don't and it falls back to the `handle` method there is no value being added by this package because a `handle` method is not likely to be compatible with the Pipeline expectations.

So while this may not be an obvious match for the style and mental model here, I think there is a reasonable justification to consider the deviation in order to add more value and convenience.

At the end of the day, I am very new to this package and if you feel strongly about moving this logic to the decorator - or somewhere else 🤔 - I'm all ears.

lorisleiva (changes requested):
I think there is a misalignment on how this package handles `AsX` decorators and this PR.

You are saying that, in order to make the `handle` method "compatible" with Pipeline one would _have to_ do the following:

```php
class PublishPost
{
    use AsAction;

    public function handle(Post $post, ?Closure $next = null): mixed
    {
        // Do the things that publish the post...

        if ($next) {
            return $next($post);
        }
    }
}
```

I disagree.

To illustrate this, let's see how `AsController` handle this situation because, it too, has some specific requirements — e.g. route model bindings.

Here's a simple example that uses an action as a controller to publish a post.

```php
class PublishPost
{
    use AsAction;

    public function handle(Post $post, PublishPostRequest $request): Response
    {
        $request->validate();
        // Do the things that publish the post...
        return redirect()->route('posts.published', [$post]);
    }
}
```

This works but because we're using the `handle` function, we are "locking" ourselves to only use this action as a controller. Sometimes that's what we want, but most of the time, you want to extract the "main logic" of your action in the `handle` method. That way, you can use `asX` functions as adapters to plug in your action in various places in the framework.

To improve on the previous design, we would need to use the `handle` method exclusively for the "domain logic" of our action, and refer to it inside the `asController` method whose responsibility is purely to make this action accessible in web/api routes. Here's our updated example.

```php
class PublishPost
{
    use AsAction;

    public function handle(Post $post): void
    {
        // Do the things that publish the post...
    }

    public function asController(Post $post, PublishPostRequest $request): Response
    {
        $request->validate();
        $this->handle($post); // Notice how we delegate to the domain logic here.
        return redirect()->route('posts.published', [$post]);
    }
}
```

Now, with that in mind, how would you make this action _also_ available as a pipeline? With your current design, we cannot do it anymore because you are forcing the `handle` method to be pipeline-specific. However, if we follow the same pattern as controllers, we can add a new `asPipeline` method that again delegates the `handle` method for the domain logic.

Which gives us the following code:


```php
class PublishPost
{
    use AsAction;

    public function handle(Post $post): void
    {
        // Do the things that publish the post...
    }

    public function asController(Post $post, PublishPostRequest $request): Response
    {
        $request->validate();
        $this->handle($post);
        return redirect()->route('posts.published', [$post]);
    }

    public function asPipeline(Post $post, ?Closure $next = null): mixed
    {
        $this->handle($post);
        if ($next) {
            return $next($post);
        }
    }
}
```

And just like that an action can be used as a pipeline and as whatever else the user needs. And if the user decides they only want this action as a pipeline, they can use the `handle` method as a fallback for the pipeline decorator (just like we did initially with the controller) and we end up with your original design which also works.

This is my main design change request on this PR. I wouldn't feel confortable merging something that deviates from the mental model of the whole package.

An additional smaller concern I have is the signature `(Post $post, ?Closure $next = null): mixed`. Particularly the input/output selection of the pipeline. If you could add more tests showing that a single action can accept multiple different inputs and return multiple different outputs without the usage of `mixed`, I'd feel more confortable with the function signature. 🙏

stevenmaguire:
@lorisleiva thank you very much for the detailed and thorough feedback. I appreciate it! I think we can make something work here, especially within the constraints of the explanation you provided.

I've pushed up an update that - I think - puts this closer to what you might consider acceptable. I'm not afraid of another round of refinement if needed.

I gleaned a few of assumed truths from your feedback that were most productive in the latest revision. I want to put them next to numbers so it might be easier to confront and/or refine each assumed truth independently, if needed.

1. It is OK to include Pipeline closure resolution logic in this package.
2. Each "critical" method in the Action (`handle` and `asPipeline` in this case) should be expected to declare unambiguous type hinting for parameters and return types.
3. Much like the `AsController` flow, the `asPipeline` method should be used to "coerce" a single Pipeline parameter into the parameters expected by the `handle` method if the `handle` method is not already Pipeline compatible.
4. If the `handle` method is Pipeline compatible (in that it only requires one non-optional parameter to function properly) it is not essential for the Action to furnish an `asPipeline` method.

I did add more test cases here to try to exercise these assumed truths relative to the Pipeline concern. Here is the Pest output for that:

```shell
   PASS  Tests\AsPipelineTest
  ✓ it can run as a pipe in a pipeline, with explicit trait
  ✓ it can run as a pipe in a pipeline, with implicit trait
  ✓ it can run as a pipe in a pipeline, without an explicit asPipeline method
  ✓ it it can run as a noop/passthrough pipe in a pipeline, without a handle or asPipeline method
  ✓ it it can run with an arbitrary via method configured on Pipeline
  ✓ it cannot run as a pipe in a pipeline, with an explicit asPipeline method expecting multiple non-optional params
  ✓ it cannot run as a pipe in a pipeline, without an explicit asPipeline method and multiple non-optional handle params
```

Hopefully between those Pest test cases and the assumed truths listed above, the current state of the PR is more clearly aligned. Please do let me know if I am still missing something.

src/Decorators/PipelineDecorator.php
```diff
@@ -0,0 +1,60 @@
...
+     *         be supplied to the next closure.
+     */
+    protected function handleFromAnyMethod(mixed ...$arguments): mixed
+    {
+        $passable = array_shift($arguments);
+        $closure = array_pop($arguments);
+        $returned = null;
+
+        if ($this->hasMethod('asPipeline')) {
+            $returned = $this->callMethod('asPipeline', [$passable]);
```
edalzell, line 53: 100% personal preference here but I'd prefer:

```php
$returned = $this->callMethod('asPipeline', [$passable]);
return $closure($returned ?? $passable);
```

Here to remove the `else`

edalzell:
There we go, great work @stevenmaguire, that's what I meant by "copy the patterns" in a package. Now it looks like all the other action types.

stevenmaguire:
@lorisleiva do you have any further feedback here given the latest changes?

int3hh:
why not merge ? cool feature to have

edalzell:
> why not merge ? cool feature to have

You can always composer patch this in if you want it now.

it-can:
really need this, any eta?

src/ActionManager.php
```diff
@@ -73,7 +73,7 @@ public function getDesignPatterns(): array
     public function registerDesignPattern(DesignPattern $designPattern): ActionManager
     {
         $this->designPatterns[] = $designPattern;
-        
+
```
Wulfheart, line 76: No reformatting in a PR on unrelated fields.

src/ActionManager.php
```diff
@@ -136,7 +136,7 @@ public function identifyFromBacktrace($usedTraits, ?BacktraceFrame &$frame = nul
         $designPatterns = $this->getDesignPatternsMatching($usedTraits);
         $backtraceOptions = DEBUG_BACKTRACE_PROVIDE_OBJECT
             | DEBUG_BACKTRACE_IGNORE_ARGS;
-        
+
```
Wulfheart, line 139: No reformatting in a PR on unrelated fields.

Wulfheart (changes requested):

Wulfheart:
> really need this, any eta?

Due to the complexity this is a PR only @lorisleiva should merge as I am unfamiliar with it.

it-can:
> complexity

ok dude

lorisleiva:
Hey guys, I've not come back to you yet because, at first glance, there are still a few smaller design decisions I'm not 100% sure about.

I need to take the time to dig properly into the changes and see if there are a few tweak we can do to mitigate this. Rest assured, this will end up being merged as I think it is a valuable feature. However, this is a fairly large addition to an otherwise stable package and I don't want to rush this. Hope you understand.

That being said, I'll try and allocate some time this weekend for this PR.

lorisleiva (changes requested):
Okay I commented all my thoughts on the current design. It looks like a lot but actually we're much closer to merging point than I thought we were after reflection haha.

Let me know what you think about my comments and do another round of review afterwards.

src/Decorators/PipelineDecorator.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * Because of these expectations, this behavior is asserting two opinions:
+     *
+     *     1)  Regardless of the number of parameters provided to the asPipeline
+     *         method implemented here, only the first will be supplied to the
+     *         invoked Action.
+     *     2)  If the invoked Action does not return anything, then the next
+     *         closure will be supplied the same parameter. However, if the
+     *         invoked action does return a non-null value, that value will
+     *         be supplied to the next closure.
```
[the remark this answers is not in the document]
stevenmaguire: Yes, this approach was ultimately made because of assumed truth No. 2 in my previous update.

> Each "critical" method in the Action (handle and asPipeline in this case) should be expected to declare unambiguous type hinting for parameters and return types.

The question you are outlining is exactly what I was struggling with - where best to handle the closure. 

In the end, I decided on this approach - and it sounds like you are gravitating that way as well - because unlike the Route middleware, most Pipeline use cases (and even the stock logic) doesn't care much about the surrounding logic of the closure. It only wants to know what to pass into it. 

In this case, that is still in the hands of the implementation and can be type hinted via return type. If the closure is supplied to `asPipeline` as outlined in your second example (and my first run at this) then a return type on the `asPipeline` method is not feasible.

I think the current proposed approach is the best.

src/Decorators/PipelineDecorator.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function __invoke(mixed ...$arguments): mixed
+    {
+        return $this->handleFromAnyMethod(...$arguments);
+    }
+
+    public function handle(mixed ...$arguments): mixed
+    {
+        return $this->handleFromAnyMethod(...$arguments);
+    }
```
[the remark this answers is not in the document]
stevenmaguire: The pipeline infrastructure's default method is also called "handle" unless explicitly changed with the `via` method.

If the Action class has a handle method that is not compatible with Pipeline (which is VERY likely), there would be a collision if stock Laravel pipeline handling attempted to use that flavor of handle method instead of the one that supports the $passable and $next parameters.

I wanted to get in front of that possible collision. Perhaps it is irrelevant? I wrote a test specifically to detect this. I'll expand on that to poke at this issue further.

src/Decorators/PipelineDecorator.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     *     1)  Regardless of the number of parameters provided to the asPipeline
+     *         method implemented here, only the first will be supplied to the
+     *         invoked Action.
+     *     2)  If the invoked Action does not return anything, then the next
+     *         closure will be supplied the same parameter. However, if the
+     *         invoked action does return a non-null value, that value will
+     *         be supplied to the next closure.
+     */
+    protected function handleFromAnyMethod(mixed ...$arguments): mixed
```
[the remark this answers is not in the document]
stevenmaguire: I waffled on this as well and in the end decided on "following package conventions" and push the requirement that drops parameters down the code path even further. 

I am in favor of the change you proposed here, I think it is clearer.

tests/AsPipelineTest.php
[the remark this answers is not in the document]
stevenmaguire: No problem.

tests/AsPipelineTest.php
```diff
@@ -0,0 +1,210 @@
...
+}
+
+class AsPipelineMultipleParamHandleOnlyTest
+{
+    use AsAction;
+
+    public function handle(AsPipelinePassable $passable, int $foo): void
+    {
+        $passable->increment();
+    }
```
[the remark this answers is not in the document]
stevenmaguire: This int parameter is a leftover from some debugging I was doing. It can be removed.

src/Decorators/PipelineDecorator.php
stevenmaguire: There is a test case for the following scenario:

> it can run as a noop/passthrough pipe in a pipeline, without a handle or asPipeline method 

This implementation causes this test case to fail. 

So, I think the main question is - is this a valid test case to support? Do we care if the Action class does not have a handle or asPipeline method? Should it throw an exception and break the pipeline if the Action class doesn't have either method? or should we simply pass the $passable into the $next closure and not break anything?

stevenmaguire: If we want to keep that test case and not expect an exception, the following (or similar) could work a bit more cleanly:

```php
        $method = match (true) {
            $this->hasMethod('asPipeline') => 'asPipeline',
            $this->hasMethod('handle') => 'handle',
            default => null
        };

        if ($method) {
            return $closure($this->callMethod($method, [$passable]) ?? $passable);
        }

        return $closure($passable);
```

src/Decorators/PipelineDecorator.php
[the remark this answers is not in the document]
lorisleiva: > If the Action class has a handle method that is not compatible with Pipeline (which is VERY likely), there would be a collision

I'm a bit confused. This is not the user's Action class but the `PipelineDecorator` which we fully control. I'm not sure what sort of collision we're avoiding.

tests/AsPipelineTest.php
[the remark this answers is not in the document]
lorisleiva: Gotcha, perhaps this will be made clearer by the separation of test cases.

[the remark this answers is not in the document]
edalzell: let's rename that though to be more descriptive. Variable names should reflect their purpose. Long is fine

src/Decorators/PipelineDecorator.php
lorisleiva: What's the use case for using an Action as a pipeline that does nothing unconditionally?

stevenmaguire:
Thanks for the continued review and feedback here. Refining and solidifying the functional and style requirements is an important process. Please review the latest changes that should be closer in line with the evolving expectations for this feature addition.

edalzell:
This is so slick now now @stevenmaguire great work!

stevenmaguire:
Thanks for approving the CI runs. There are two failing builds and I'm scratching my head on one of them. If you have some thoughts, please chime in.

P8.4 - L^11.0 is failing with the error message: 

> Lorisleiva\Actions\Tests\AsPipelineWithMultipleNonOptionalParametersTest::handle(): Argument #2 ($nonOptionalAdditionalParameter) must be of type int, Closure given, called in /home/runner/work/laravel-actions/laravel-actions/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php on line 209

This is the concern I was mentioning in this thread of the feedback: https://github.com/lorisleiva/laravel-actions/pull/304#discussion_r1954852475

But it is curious that it is only an issue with 8.4. I can reproduce this issue locally on 8.4. Laravel's Pipeline handling logic is detecting the handle method on the Action class - the one that expects two parameters in this failing test case - and is invoking that method instead of the __invoke method on the decorator. Furnishing a handle method on the decorator (like a previous iteration of this solution) does not change this behavior. 

After poking at it for a while, it seems - and I could be wrong - that the `$pipe` in consideration within Laravel's Pipeline logic may start out as an instance of the PipelineDecorator eventually resolves as the direct action instance, which has its own handle method - skipping the decorator logic.

P8.4 - L^10.0 is failing with the error message: 

> Class "Illuminate\Support\Facades\Pipeline" not found

This one is a head scratcher. Could there be something wrong with the autoloading of the Laravel framework in this specific environment?

stevenmaguire:
It seems that the Backtrace frame matching is different between <8.4 and 8.4. 

Below is the value of `$frame->function` for the various environments:

```
26-02-2025 17:10:10 @ 8.2.27 - Illuminate\Pipeline\{closure} 
26-02-2025 17:10:19 @ 8.3.17 - Illuminate\Pipeline\{closure} 
26-02-2025 17:10:28 @ 8.4.4 - {closure:{closure:Illuminate\Pipeline\Pipeline::carry():184}:185} 
```

Adding this new pattern to the frame matching gets over that hump but is a little smelly.

A couple of other observations from this debugging...

1) The `PipelineDecorator` needs to ensure that something is always returned when the `$closure` returns null.
2) The Pipeline stack cannot include a standalone instance of an Action (eg: `new PipelineCompatibleAction` as opposed to `PipelineCompatibleAction::class`). There is a test case demonstrating this.
3) The Pipeline stack can include only one explicit container resolved instance (eg: `app()->make(PipelineCompatibleAction::class)`) at the bottom of the stack. There is a test case demonstrating this.

New changes are pushed up and ready for review and CI.

Regarding #3 above, this feels like a ServiceProvider timing issue that _could_ be overcome (maybe?) but I am not super familiar with that area of the package yet. Perhaps there are some ideas on whether or not that is something we should expect to overcome? Or we can simply assert that this behavior is ONLY expected to work when using the `PipelineCompatibleAction::class` approach when building the pipeline stack?

To spare you digging into the test cases, here is an example of what works perfectly:

```php
it('can run as a pipe in a pipeline, with an explicit asPipeline method', function () {
    $passable = Pipeline::send(new PipelinePassable)
        ->through([
            AsPipelineTest::class,
            AsPipelineTest::class,
            AsPipelineTest::class,
            AsPipelineTest::class,
        ])
        ->thenReturn();

    expect(is_a($passable, PipelinePassable::class))->toBe(true);
    expect($passable->count)->toBe(4);
});
```

Here is an example of what gives the impression of working but actually is not:

```php
it('can run as a pipe in a pipeline with only one explicit container resolved instance at the bottom of the stack', function () {
    $passable = Pipeline::send(new PipelinePassable)
        ->through([
            AsPipelineTest::class, // implicit container resolved instance
            app()->make(AsPipelineTest::class), // explicit container resolved instance
        ])
        ->thenReturn();

    expect(is_a($passable, PipelinePassable::class))->toBe(true);
    expect($passable->count)->toBe(2);
});
```

Here are two examples of failing behavior:
```php
it('cannot run as a pipe in a pipeline with an explicit container resolved instance in the middle of the stack', function () {
    $passable = Pipeline::send(new PipelinePassable)
        ->through([
            AsPipelineTest::class, // implicit container resolved instance
            app()->make(AsPipelineTest::class), // explicit container resolved instance
            AsPipelineTest::class, // implicit container resolved instance
            AsPipelineTest::class, // implicit container resolved instance
        ])
        ->thenReturn();

    expect(is_a($passable, PipelinePassable::class))->toBe(true);
    expect($passable->count)->toBe(2); // 4 is the ideal count here
})->only();

it('cannot run as a pipe in a pipeline as an standalone instance', function () {
    $passable = Pipeline::send(new PipelinePassable)
        ->through([
            new AsPipelineTest, // standalone instance
            AsPipelineTest::class, // implicit container resolved instance
            app()->make(AsPipelineTest::class), // explicit container resolved instance
        ])
        ->thenReturn();

    expect(is_null($passable))->toBe(true);
});
```

stevenmaguire:
@lorisleiva do you have a thought here on the fork in the road given the test case support mentioned above? 

Should we: 

1. Set an expectation that this package will only support pipeline pipes configured as using the `::class` syntax so the framework resolves the class consistently where the frame matching logic expects, or
2. Explore the (seemingly remote) possibility of expanding the frame matching logic to support pipeline pipes configured as instances?

I think option 1 is the path of least resistance and will undoubtedly need some careful hand-holding in the documentation to effectively manage the expectation that if any pipeline enabled Action is used, it must follow the convention.

lorisleiva:
Hi @stevenmaguire, Sorry for the late reply. I needed time to digest this properly.

Firstly, I wanted to mention that we cannot realistically rely on line numbers when matching frames. These numbers (184 and 185 in the snippet below) will update every time some code gets added or deleted in the `Pipeline` class which is not a robust solution and will be a nightmare to maintain as we always support more than one Laravel version at a time.

```php
    public function recognizeFrame(BacktraceFrame $frame): bool
    {
        return $frame->matches(Pipeline::class, 'Illuminate\Pipeline\{closure}')
            || $frame->matches(Pipeline::class, '{closure:{closure:Illuminate\Pipeline\Pipeline::carry():184}:185}');
    }
```

Whilst we could use a regex to fix this, it is time to take a step back and compare the benefits of this PR with the cognitive and maintenance overhead it brings.

**Option 1**

Your first option (only supporting the `::class` API when registering actions as pipelines) bring the following benefit to users.

```diff
    $passable = Pipeline::send(new PipelinePassable)
-       ->via('asPipeline')
        ->through([MyAction::class])
        ->thenReturn();
```

And the benefit of not having to define the `asPipeline` function _if and only if_ they are planning on sticking to a simple "passable in / passable out" kind of pipeline.

The cons of this approach — on top of the maintaining overheads — is that it creates expectations for users that may not be met. What if I need a custom `asPipeline` method that changes the way the `next` closure is called? In this case, implementing the `asPipeline` function on the Action will require a function signature that differs from what I would expect from the framework.

**Option 2**

Other expectations include the problems you are trying to solve in option 2 such as invoking a pipeline via `app()->resolve()` or `new MyAction`.

Solving these problems with complex backtrace frame matching and more hooks on the IoC container _may_ be possible but _should_ we?

By that point, we would have made this library much more difficult to maintain — and much more prone to errors as the framework evolves — for a Nice To Have feature. Which leads me to...

**Option 3**

We do nothing.

If people wanted to improve the developer experience around pipelines for their actions, they'd just need to create a `AsPipeline` trait with a solution akin to the one [provided here](https://github.com/lorisleiva/laravel-actions/issues/279#issuecomment-2022041507) and just use the `via("asPipeline")` modifier. This seems like a much more user-friendly solution than having to teach users a whole new way of handling Pipeline that is kinda like the framework but also kinda not.

Right now, I am heavily leaning into option 3.

> Rest assured, this will end up being merged as I think it is a valuable feature.

I'm sorry for having committed to this too early but, unfortunately, it turns out this innocent little feature brings a lot more baggage than it deserves.

I'll leave this open for a while in case I'm missing something or people want to change my mind but please bear in mind that this sort of things could just as well exist in separate packages.

stevenmaguire:
Thanks @lorisleiva. That is exactly where I wanted to head in this PR - "is the juice worth the squeeze?" 

It was clear from the original Issue and my early commits on the PR that there was not an obvious and easy "drop-in" solution available. I have centered my focus here on finding all the places where this could go wrong so that either 1) someone else with more domain knowledge (you and others) might highlight some kind of "ah-ha" moment to make it click or 2) make sure there is an informed reason why this is NOT a good idea to implement. I think the latter, No. 2, is where we're at based on your comments and I think that is OK.

I do think this support would be valuable and wish it were more viable at this point.

lorisleiva:
I understand. It is also frustrating for me not to be able to help you push this over the line. I wished I’d been able to identify these issues earlier to avoid you spending as much time on this.

Perhaps this is something we can reevaluate in later versions of the framework where at least there won’t be any backtrack frame discrepancies between PHP versions anymore.

edalzell:
Hot take I know, but you could ONLY support L12 & php 8.4 if you wanted. If folks don't want to upgrade they won't get this feature. i.e. make it a breaking change.

lorisleiva:
Only supporting L12/PHP8.4 would only simplify the backtrace frame matching for using `AsPipelineTest::class` in Pipelines. It won't change the fact that `new AsPipelineTest` and `app()->resolve(AsPipelineTest::class)` will cause unexpected behaviour to the user. If fact, since Laravel Actions works by recognising and decorating special classes when they are resolved from the container, there wouldn't be a way of changing this unless we completely overwrote the `Pipeline` class itself.

The only way forward I can see for this PR is:
- We implement a more robust backtrace frame matching on `'{closure:{closure:Illuminate\Pipeline\Pipeline::carry():184}:185}'` (e.g. we only check it contains `closure:Illuminate\Pipeline\Pipeline::carry()`).
- We do not support `new MyAction` or `app()->resolve(MyAction::class)` because there's nothing we can do to recognise this. We add one test file for each of these patterns showing they don't work and we keep the rest of the test clean with `MyAction::class` only.
- We do not add `AsPipeline` to the `AsAction` trait by default. It's opt-in. That way, we avoid adding unexpected behaviour to people's apps unless they know what they're doing.
- We add a whole new page in the documentation explaining all this. The documentation PR should be ready to merge before this gets merged.

What do you guys think of this plan?

tests/AsPipelineTest.php
```diff
@@ -0,0 +1,87 @@
...
+        ])
+        ->thenReturn();
+
+    expect(is_a($passable, PipelinePassable::class))->toBe(true);
+    expect($passable->count)->toBe(4);
+});
+
+it('can run with an arbitrary via method configured on Pipeline', function () {
+    $passable = Pipeline::send(new PipelinePassable)
+        ->via('arbitraryMethodThatDoesNotExistOnTheAction')
```
lorisleiva, line 40: Side note: I'm curious to understand why this works btw. Is it because if a pipeline class has an `__invoke` method, the framework will fallback to using this instead of `arbitraryMethodThatDoesNotExistOnTheAction`?

stevenmaguire:
> We implement a more robust backtrace frame matching on '{closure:{closure:Illuminate\Pipeline\Pipeline::carry():184}:185}' (e.g. we only check it contains closure:Illuminate\Pipeline\Pipeline::carry()).

I think adding some regex or Str::contains logic to the matcher would get over this hump. Yes it adds a bit of overhead when there are major version bumps of Laravel. I am not the best person to speak to whether or not this is too onerous from a maintenance perspective. I can add this logic to the PR as it appears to be straightforward? Also, I think there is no reason to "only" make this feature work for L12 - the previous matching rule works just fine for previous versions of Laravel and PHP.

> We do not support `new MyAction` or `app()->resolve(MyAction::class)` because there's nothing we can do to recognise this. 

Based on the amount of work required and the implied hesitation to taking this work on, I think this is the most reasonable way to approach this feature. Spelling this out as a limitation in the documentation is the prudent option, if you decide to move forward. 

> We add one test file for each of these patterns showing they don't work and we keep the rest of the test clean with MyAction::class only.

This shouldn't be an issue.

> We do not add AsPipeline to the AsAction trait by default. It's opt-in. That way, we avoid adding unexpected behaviour to people's apps unless they know what they're doing.

This is also a prudent choice to attempt to act in the developer's best interest in not introducing unexpected behavior.

> We add a whole new page in the documentation explaining all this. The documentation PR should be ready to merge before this gets merged.

Considering all of the hand-holding you proposed and mentioned above, creating some space to properly lay all of this out is, IMO, the best way forward. 

----------------------------------

I think if there is an appetite for all of these tradeoffs, especially from the maintenance and cognitive load perspective, then this is a reasonable plan to move forward. There is def a NET positive lift here in the value being added, despite introducing some tradeoffs that might be more difficult to explain passively.

lorisleiva:
Thanks @stevenmaguire! I'm happy to move forward with this plan. May I suggest the following implementation process:
1. Make the PR changes we discussed and update the tests.
2. Make sure CI passes.
3. Write a documentation PR for the `AsPipeline` trait. In the meantime, I can give this PR another thorough review.
4. I merge and publish both PRs at the same time.

stevenmaguire:
Sounds good @lorisleiva!

> Make sure CI passes.

Is there a way to update this PR or some other permissions so that CI is approved automatically? On previous commits it did not run automatically and needed to wait for approval.

PR #305, opened 2025-01-13
Fix php8.4 deprecated notices

it-can:
Fixes

```
PHP Deprecated:  Lorisleiva\Actions\Exceptions\MissingCommandSignatureException::__construct(): Implicitly marking parameter $previous as nullable is deprecated, the explicit nullable type must be used instead in ./src/Exceptions/MissingCommandSignatureException.php on line 10
```


it-can:
@adrenallen can this be merged?

Wulfheart:
Thanks.

PR #308, opened 2025-02-10
Injects route parameters into action authorize

CWAscend:
## 💪 Motivation

Currently, when running the authorize method on an Action, the only variable you can access is the ActionRequest by resolving it from the container via dependency injection. This can sometimes lead to the authorize method becoming messy if we need to access a route-model bound model, as an example, and authorize that a certain action can run based on the state of the particular model.

This example might explain it a bit better:

```php
// Current implementation
class ShowOrder
{
    use AsController;

    public function authorize(ActionRequest $request): bool
    {
        /** @var Order $order */
        $order = $request->order;

        return $order->status->is(OrderStatus::COMPLETE);
    }

    public function handle(Order $order): Response
    {
        // ...
    }
}

// New implementation
class ShowOrder
{
    use AsController;

    public function authorize(Order $order): bool
    {
        return $order->status->is(OrderStatus::COMPLETE);
    }

    public function handle(Order $order): Response
    {
        // ...
    }
}
```

The ActionRequest can still be resolved via dependency injection, so this isn't a breaking change.

## 🛠 Changes

- Pass route parameters into the `authorize` method, when authorizing the incoming request
- This allows us to tidy our authorize method, avoiding the need to pluck parameters from the request

## 🧪 Testing

- A test has been added to `tests/AsControllerWithAuthorizeAndRulesTest.php`

## 🧙 Reminders (Author)

- [x] Have you read through your own diff? 👀
- [x] Are all tests passing? 🧪

### DIFF
--- a/src/Concerns/ValidateActions.php
+++ b/src/Concerns/ValidateActions.php
@@ -133,8 +133,12 @@ trait ValidateActions
     protected function inspectAuthorization(): Response
     {
         try {
+            $routeParameters = method_exists($this, 'route') ? $this->route()->parameters() : null;
+
             $response = $this->hasMethod('authorize')
-                ? $this->resolveAndCallMethod('authorize')
+                ? ($routeParameters
+                    ? $this->resolveAndCallMethod('authorize', $routeParameters)
+                    : $this->resolveAndCallMethod('authorize'))
                 : true;
         } catch (AuthorizationException $e) {
             return $e->toResponse();
--- a/tests/AsControllerWithAuthorizeAndRulesTest.php
+++ b/tests/AsControllerWithAuthorizeAndRulesTest.php
@@ -32,9 +32,25 @@ class AsControllerWithAuthorizeAndRulesTest
     }
 }
 
+class AsControllerWithAuthorizeBindingsTest
+{
+    use AsController;
+
+    public function authorize(string $someRouteParameter): bool
+    {
+        return $someRouteParameter !== 'unauthorized';
+    }
+
+    public function handle(ActionRequest $request, string $someRouteParameter)
+    {
+        return [$someRouteParameter];
+    }
+}
+
 beforeEach(function () {
     // Given the action is registered as a controller.
     Route::post('/calculator', AsControllerWithAuthorizeAndRulesTest::class);
+    Route::get('/authorize-bindings/{someRouteParameter}', AsControllerWithAuthorizeBindingsTest::class);
 });
 
 it('passes authorization and validation', function () {
@@ -85,4 +101,9 @@ it('uses a new validator at every request', function () {
 
     $this->post('/calculator', ['operation' => 'invalid'])
         ->assertSessionHasErrors(['operation', 'right', 'left']);
+});
+
+it('resolves route parameters as authorization arguments', function () {
+    $this->get('/authorize-bindings/authorized')->assertOk()->assertExactJson(['authorized']);
+    $this->get('/authorize-bindings/unauthorized')->assertForbidden();
 });


CWAscend:
Might be worth doing the same here as well, but will leave that open for discussion 

https://github.com/lorisleiva/laravel-actions/blob/main/src/Concerns/ValidateActions.php#L85-L90

CWAscend:
@lorisleiva hey mate, any idea if this is going to be merged or not?

lorisleiva:
Hey! Thanks for the reminder. I think that makes sense. Am I right to say that this is what would be passed in the `authorize` method if you were not using actions? Just want to make sure devs would expect these parameters to be injected as, once we pass them in, it would be a breaking change to change that behaviour.

adrenallen:
Ran into this today, would love this functionality. Currently if you define a model as the param, the authorize will instantiate an empty version of that model which honestly confused me at first and led to a bug.

Having authorize inject params the same way controller does is what I expected the functionality to be out of the box. This PR would be great and seems relatively low impact.

i.e. **this will log a document model, but it will be empty 😓**
image


At the moment I'm using this approach to get the route params, which is a bit messy and differs from other in-signature references for the rest of the action

image

adrenallen:
@lorisleiva bump on this

CWAscend:
> Hey! Thanks for the reminder. I think that makes sense. Am I right to say that this is what would be passed in the `authorize` method if you were not using actions? Just want to make sure devs would expect these parameters to be injected as, once we pass them in, it would be a breaking change to change that behaviour.

Yes, exactly this :)

vpratfr:
@lorisleiva Hi, I'd like to tackle that PR next. Would you be able to allow me launching pipelines too? I would like it to get checked on PHPStan & co.

vpratfr (commented):
@CWAscend I have asked a review to Claude and here it is. Looks like there are a few things to address before I can comfortably merge it.

---

I read the diff against `main` and checked the actual resolution mechanics in Laravel 12.64 (`BoundMethod` vs `ResolvesRouteDependencies`). The feature is very welcome; a few points below that I think are worth settling before merge, plus an answer to @lorisleiva's BC question.

## 1. Answering the BC question

> Am I right to say that this is what would be passed in the `authorize` method if you were not using actions?

Not quite. In a plain `FormRequest`, `passesAuthorization()` is:

```php
$result = $this->container->call([$this, 'authorize']);
```

No extra arguments — so vanilla Laravel does **not** inject route parameters into `authorize()` either, and a model-typed parameter there gives you the same empty model @adrenallen ran into. The consistency this PR really buys is with **the action's own `handle()` / controller signature**, not with `FormRequest`. That's a stronger argument, and it matches what people expect from an action.

On the BC surface itself, it looks small but not empty:

- Behaviour only changes for an `authorize` parameter whose **name** matches a route parameter name. Previously those cases produced either an empty model or a `BindingResolutionException`, so nobody can reasonably depend on the old outcome.
- One silent exception: `authorize(ActionRequest $request)` on a route containing `{request}`. `BoundMethod::addDependencyForCallParameter()` checks the parameter **name** before the type, so `$request` silently becomes a string. Rare, but silent.
- Unconsumed route parameters are appended **positionally** (`array_merge($dependencies, array_values($parameters))` in `getMethodDependencies()`). Harmless in PHP (extra args are ignored), but it does mean every existing action's `authorize` now receives surplus arguments.

## 2. `route()` can be null, and `parameters()` keeps nulls

`Request::route()` is typed `Route|object|string|null`. Through the library's own flow (`ControllerDecorator::__invoke()` → `$request->validate()`) a route is always resolved, so `->parameters()` on null isn't reachable today — but it's free to guard.

More importantly, `ControllerDecorator::resolveFromRouteAndCall()` uses `parametersWithoutNulls()`, not `parameters()`. With `parameters()`, a route like `/{slug?}` called without the segment injects `null` into `authorize(string $slug = 'x')` instead of letting the default apply. Worth aligning.

## 3. `method_exists($this, ...)` and the nested ternary

The trait introspecting itself to tell `ActionRequest` from `AttributeValidator` is a bit of a smell. An overridable hook reads better and removes the nested ternary — note that `resolveAndCallMethod($m, [])` is exactly equivalent to `resolveAndCallMethod($m)`, so the `$routeParameters ? … : …` branch is redundant:

```php
// ValidateActions.php
$response = $this->hasMethod('authorize')
    ? $this->resolveAndCallMethod('authorize', $this->getAuthorizationArguments())
    : true;

/** Extra arguments injected into the action's authorize method, by name. */
protected function getAuthorizationArguments(): array
{
    return [];
}

// ActionRequest.php
protected function getAuthorizationArguments(): array
{
    $route = $this->route();

    return $route instanceof Route ? $route->parametersWithoutNulls() : [];
}
```

## 4. Injection semantics diverge from `handle()`

This is the real design decision. `handle()` goes through `resolveClassMethodDependencies()`: class-typed parameters are matched **by type** (route bindings) and the rest **positionally**. `app()->call()` (via `resolveAndCallMethod`) matches **by name** first. Concretely:

```php
// route: /orders/{order}
public function authorize(Order $theOrder) { … } // ❌ name != 'order' -> container->make(Order) -> EMPTY model
public function handle(Order $theOrder) { … }    // ✅ works (matched by type)
```

So the PR doesn't remove the empty-model footgun, it moves it onto name matching. Either document explicitly that the parameter name must match the route parameter name, or reuse `ResolvesRouteDependencies` for true parity with `handle()`. Name-based matching is defensible (arguably more predictable than positional), it just deserves to be a stated choice.

## 5. `WithAttributes` actions are not covered

When an action uses `WithAttributes`, `ControllerDecorator::shouldValidateRequest()` returns `false` and validation runs through `AttributeValidator`, which has no `route()` — so no injection happens there. `authorize()`'s behaviour would then depend on which trait the action uses. Either close the gap (`ControllerDecorator` already holds `$this->route`) or document the limitation.

## 6. Tests

The added test only covers a `string` parameter. In rough order of value, I'd add:

- an actual **route-model binding** — it's the motivating use case, and the only test that proves the empty model is gone;
- a regression test for `authorize(ActionRequest $request)` **on a route that has parameters**, proving the surplus positional args break nothing;
- a missing optional route parameter;
- the `WithAttributes` path, if only to pin the expected behaviour.

Minor convention note: the helper class `AsControllerWithAuthorizeBindingsTest` lives inside another Pest file, whereas the repo follows "one file, one action named after the file". A dedicated `AsControllerWithAuthorizeBindingsTest.php` would fit better.

## 7. Misc

- **No CI has run on this PR** (empty status checks — workflows presumably awaiting maintainer approval). Worth noting PHPStan runs at level 5, and `$this->route()->parameters()` on a `Route|object|string|null` union looks likely to be flagged; the `instanceof Route` above takes care of it.
- Docs live in a separate repo, so an update mentioning the parameter-naming rule would be needed.
- On your own open question about doing the same for `rules()` (L85-90): I'd keep this PR to `authorize` only. Generalising touches `rules`/`messages`/`withValidator`/`afterValidator`/`getValidator` and introduces **key collisions** with the existing `compact('validator')` / `compact('factory')` calls — a route parameter named `validator` would overwrite the validator. Probably a separate issue.

lorisleiva:
@vpratfr I can't run the pipeline either haha. I think the PR is too old now and GitHub says no. Sorry. 🙈

vpratfr:
closed for https://github.com/lorisleiva/laravel-actions/pull/348 so we can work on it

Issue #313, opened 2025-03-12
Typed property Lorisleiva\Actions\ActionRequest::$action must not be accessed before initialization with knuckleswtf/scribe:5.1 when using ActionRequest


FaizAhmadSE:
Getting this error with knuckleswtf/scribe:5.1 when using ActionRequest

Typed property Lorisleiva\Actions\ActionRequest::$action must not be accessed before initialization
at vendor/lorisleiva/laravel-actions/src/Concerns/DecorateActions.php:33
     29▕     }
     30▕ 
     31▕     protected function hasMethod(string $method): bool
     32▕     {
  ➜  33▕         return method_exists($this->action, $method);
     34▕     }
     35▕ 
     36▕     protected function callMethod(string $method, array $parameters = [])
     37▕     {

jeffersonmartin:
I'm seeing this after installing Scramble docs generator too.

```
"dedoc/scramble-pro": "^0.7",
"dedoc/scramble": "^0.12",
```

```
⚠️ Cannot generate request documentation: Typed property Lorisleiva\Actions\ActionRequest::$action must not be accessed before initialization
```

Wulfheart:
Seems like an issue with scramble to me.

PR #320, opened 2025-05-01
Initialize $action property and guard hasMethod()

behzadsp:
When using ActionRequest under PHP 7.4+, calling method_exists($this->action, …) before $action has been initialized causes a fatal error:

    Typed property Lorisleiva\Actions\ActionRequest::$action must not be accessed before initialization
    
[Issue #313 ](https://github.com/lorisleiva/laravel-actions/issues/313)

This happens because typed properties in PHP 7.4+ are in an uninitialized state unless given a default value
[php.net](https://www.php.net/manual/en/language.oop5.properties.php)

Solution

    Initialize the $action property to null
    Ensures the property is always in a defined state.

    Guard the hasMethod() check with isset($this->action)
    Prevents calling method_exists() on an uninitialized property.

Testing

    ./vendor/bin/pest --filter=DecorateActionsTest

### DIFF
--- a/src/Concerns/DecorateActions.php
+++ b/src/Concerns/DecorateActions.php
@@ -4,7 +4,7 @@ namespace Lorisleiva\Actions\Concerns;
 
 trait DecorateActions
 {
-    protected mixed $action;
+    protected mixed $action = null;
 
     public function setAction($action): self
     {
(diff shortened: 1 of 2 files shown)


ceejayoz:
I can confirm this addresses an issue I'm seeing with Scribe (running `php artisan scribe:generate`).

lorisleiva:
This package no longer supports PHP 7.x. Is this issue occurring under the supported PHP versions?

ceejayoz:
I'm using PHP 8.4 at this time.

lorisleiva (approved):
Gotcha, thanks for the clarification. 🙏

PR #338, opened 2026-03-24
Add PHPStan extension

### LINKED ISSUE #260: PHP Stan support
It doesn't throw errors in PHPStan if args type, numbers etc. mismatches...
```php
namespace App\Authentication\Actions;

use Lorisleiva\Actions\Concerns\AsAction;

class UpdateUserPassword
{
    use AsAction;

    public function handle(User $user, string $newPassword)
    {
        $user->password = Hash::make($newPassword);
        $user->save();
    }
}
```
```
// Equivalent to "UpdateUserPassword::make()->handle($user, 'secret')".
UpdateUserPassword::run($user, 'secret');
```

If I use this - UpdateUserPassword::make()->handle($user, 'secret')
Larstan properly check args type, count etc

For this - UpdateUserPassword::run($user, 'secret')
Larstand doesn't trigger any error 

How can I add support to this? I am open to creating a PR. A little head start will be good...

pindab0ter:
## Summary

Closes #260

`AsObject::run()` is defined as `(mixed...): mixed`, so PHPStan treats it as untyped: no return type inference, no parameter validation, and `@throws` annotations on `handle()` are invisible. This adds a PHPStan extension that makes `run()`, `runIf()`, and `runUnless()` transparent proxies to `handle()` for static analysis:

- **Return types**: `Action::run()` returns whatever `handle()` returns. `runIf()`/`runUnless()` return `handle() | Fluent`.
- **Parameter validation**: argument types, count, optionality, and variadics are checked against `handle()`'s signature.
- **`@throws` propagation**: `handle()`'s declared throw types carry through to callers of `run()`/`runIf()`/`runUnless()`.

This covers `run()`, `runIf()`, and `runUnless()` only. Several other methods (job and controller related) have the same `mixed ...$arguments` gap. I'd prefer to merge this as a first step and tackle those in a follow-up, but I'm also open to extending this PR if preferred.

### Setup

Similar to extensions shipped by [Larastan](https://github.com/larastan/larastan), [Carbon](https://github.com/CarbonPHP/carbon), and [Pest](https://github.com/pestphp/pest-plugin-type-coverage): if you use [`phpstan/extension-installer`](https://github.com/phpstan/extension-installer), the extension is discovered automatically. Otherwise, include it manually in your `phpstan.neon`:

```neon
includes:
    - vendor/lorisleiva/laravel-actions/extension.neon
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     "require-dev": {
         "orchestra/testbench": "^9.0|^10.0|^11.0",
         "pestphp/pest": "^3.0|^4.0",
+        "phpstan/phpstan": "^2.1",
         "phpunit/phpunit": "^11.5|^12.0"
     },
     "autoload": {
@@ -39,7 +40,10 @@
     "autoload-dev": {
         "psr-4": {
             "Lorisleiva\\Actions\\Tests\\": "tests"
-        }
+        },
+        "classmap": [
+            "tests/PHPStan/Fixtures"
+        ]
     },
     "scripts": {
         "test": "vendor/bin/pest --colors=always",
@@ -59,6 +63,11 @@
             "aliases": {
                 "Action": "Lorisleiva\\Actions\\Facades\\Actions"
             }
+        },
+        "phpstan": {
+            "includes": [
+                "extension.neon"
+            ]
         }
     },
     "minimum-stability": "dev",
--- a/src/PHPStan/ActionHelper.php
+++ b/src/PHPStan/ActionHelper.php
@@ -30,7 +30,7 @@ final class ActionHelper
         if ($call->class instanceof Name) {
             $callerType = $scope->resolveTypeByName($call->class);
         } else {
-            $callerType = $scope->getType($call->class);
+            $callerType = $scope->getType($call->class)->getObjectTypeOrClassStringObjectType();
         }
 
         $classReflections = $callerType->getObjectClassReflections();
--- a/src/PHPStan/ActionHelper.php
+++ b/src/PHPStan/ActionHelper.php
@@ -59,7 +59,7 @@ final class ActionHelper
     }
 
     /**
-     * Removes the leading $boolean condition argument from a runIf()/runUnless() call,
+     * Removes the $boolean condition argument from a runIf()/runUnless() call,
      * so the remaining args line up with handle()'s parameters.
      *
      * @param array<Arg> $args
@@ -67,26 +67,50 @@ final class ActionHelper
      */
     public function stripConditionArg(array $args, string $methodName): array
     {
-        if (! ($methodName === 'runIf' || $methodName === 'runUnless')) {
+        $index = $this->findConditionArgIndex($args, $methodName);
+
+        if ($index === null) {
             return $args;
+        }
+
+        unset($args[$index]);
+
+        return array_values($args);
+    }
+
+    /**
+     * Finds the $boolean condition argument of a runIf()/runUnless() call.
+     *
+     * @param array<Arg> $args
+     */
+    public function getConditionArg(array $args, string $methodName): ?Arg
+    {
+        $index = $this->findConditionArgIndex($args, $methodName);
+
+        return $index === null ? null : $args[$index];
+    }
+
+    /** @param array<Arg> $args */
+    private function findConditionArgIndex(array $args, string $methodName): ?int
+    {
+        if (! ($methodName === 'runIf' || $methodName === 'runUnless')) {
+            return null;
         }
 
         if (count($args) === 0) {
-            return $args;
+            return null;
         }
 
         if ($args[0]->name === null) {
-            return array_slice($args, 1);
+            return 0;
         }
 
         foreach ($args as $i => $arg) {
             if ($arg->name !== null && $arg->name->toString() === 'boolean') {
-                unset($args[$i]);
-
-                return array_values($args);
+                return $i;
             }
         }
 
-        return $args;
+        return null;
     }
 }
--- a/src/PHPStan/ActionHelper.php
+++ b/src/PHPStan/ActionHelper.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Lorisleiva\Actions\PHPStan;
+
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\ExtendedMethodReflection;
+
+final class ActionHelper
+{
+    private const AS_OBJECT_TRAIT = 'Lorisleiva\Actions\Concerns\AsObject';
+
+    private const PROXY_METHODS = ['run', 'runIf', 'runUnless'];
+
+    public function resolveActionClass(StaticCall $call, Scope $scope): ?ClassReflection
+    {
+        if (! $call->name instanceof Identifier) {
+            return null;
+        }
+
+        if (! in_array($call->name->toString(), self::PROXY_METHODS, true)) {
+            return null;
+        }
+
+        if ($call->class instanceof Name) {
+            $callerType = $scope->resolveTypeByName($call->class);
+        } else {
+            $callerType = $scope->getType($call->class);
+        }
+
+        $classReflections = $callerType->getObjectClassReflections();
+
+        if (count($classReflections) !== 1) {
+            return null;
+        }
+
+        $classReflection = $classReflections[0];
+
+        if (! $classReflection->hasTraitUse(self::AS_OBJECT_TRAIT)) {
+            return null;
+        }
+
+        return $classReflection;
+    }
+
+    public function getHandleMethod(ClassReflection $classReflection): ?ExtendedMethodReflection
+    {
+        if (! $classReflection->hasMethod('handle')) {
+            return null;
+        }
+
+        return $classReflection->getNativeMethod('handle');
+    }
+}
--- a/src/PHPStan/ActionHelper.php
+++ b/src/PHPStan/ActionHelper.php
@@ -65,6 +65,15 @@ final class ActionHelper
         return $classReflection->getNativeMethod('handle');
     }
 
+    /**
+     * static:: resolves to the declaring class, losing the subclass that will
+     * actually receive the call.
+     */
+    public function isLateStaticBinding(StaticCall $call): bool
+    {
+        return $call->class instanceof Name && $call->class->toLowerString() === 'static';
+    }
+
     /** @param array<Arg> $args */
     public function hasUnpackedArg(array $args): bool
     {
--- a/src/PHPStan/ActionHelper.php
+++ b/src/PHPStan/ActionHelper.php
@@ -65,6 +65,18 @@ final class ActionHelper
         return $classReflection->getNativeMethod('handle');
     }
 
+    /** @param array<Arg> $args */
+    public function hasUnpackedArg(array $args): bool
+    {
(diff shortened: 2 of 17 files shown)


vpratfr:
That would be very very cool @lorisleiva 

Taking this opportunity to ask you if you need maintainers for the package. I see a couple of PR still open after more than a year, wondering if I could help on the matter.

pindab0ter:
@lorisleiva any chance you could take a look at this PR and/or at @vpratfr's offer?

lorisleiva:
Hey guys, sorry about the (very) late response. Yeah I don't keep up enough with the Laravel updates these days to be able to quickly context switch into this world so that tends to delay things significantly. Having help on the maintenance side would definitely be great! I'm mainly looking to keep the status quo (no crazy API changes or anything like that) whilst making sure the code base doesn't decay as new Laravel versions pile up.

@vpratfr is that something you'd like to tackle?

As for this PR, can I just confirm that this is not a breaking change and just something people can opt-in to have better static analysis of their code?

vpratfr:
I cannot promise I'll be ultra reactive but surely will be ok to give help.

I am all fine with keeping the library more or less like it is, improving small things and fixing others. 

We currently use the library on a rather large project, and see a few things to tackle.

I can have a look at existing MRs and see what could be nice to get in the next release. 

One that comes to my mind would be supporting route parameter injection in the authorize method.

Let me know if you want me as maintainer. Bye.

lorisleiva:
I'm sure even if you answer a few issues/PRs every now and then, it'll be better than the current status quo. I'll send you an invite then.

As for this PR, let me just invoke my AI reviewer first to see if there's anything relevant it would find.

composer.json
```diff
@@ -39,7 +40,10 @@
     "autoload-dev": {
         "psr-4": {
             "Lorisleiva\\Actions\\Tests\\": "tests"
-        }
+        },
+        "classmap": [
+            "tests/PHPStan/Fixtures"
+        ]
```
trevor-cortex, line 46: Two thoughts on classmapping the fixtures:

1. Is this actually required? PHPStan's `RuleTestCase`/`TypeInferenceTestCase` analyse the fixture files directly with static reflection, so the classes defined in them are usually discoverable without being autoloadable. If the tests pass without it, I'd drop it.
2. If it is needed, note that these fixture files contain top-level executable code (`AddAction::run(1, 2);` etc.). Classmap generation only parses, but if anything ever triggers a runtime autoload of one of these classes, the whole file executes — including `app(static::class)` calls that will blow up outside a booted container. A comment or a defensive `if (false)`-style guard isn't worth it, but it's a footgun to be aware of if the fixtures ever get referenced from a Pest test.

src/PHPStan/ActionHelper.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if (! in_array($call->name->toString(), self::PROXY_METHODS, true)) {
+            return null;
+        }
+
+        if ($call->class instanceof Name) {
+            $callerType = $scope->resolveTypeByName($call->class);
+        } else {
+            $callerType = $scope->getType($call->class);
+        }
```
trevor-cortex, line 41: Minor gap: when the callee is a variable holding a class-string (e.g. `$action::run(...)` where `$action` is `class-string<AddAction>`), `$scope->getType($call->class)` yields a `ClassStringType`, whose `getObjectClassReflections()` is empty — so the extension silently bails and the call stays `mixed`.

If you want to cover that pattern, `$scope->getType($call->class)->getObjectTypeOrClassStringObjectType()` handles both the object and class-string cases in one call. Non-blocking — the common `SomeAction::run()` form is what matters most.

src/PHPStan/RunParameterValidationRule.php
```diff
@@ -0,0 +1,169 @@
...
+                ))
+                    ->identifier('laravelActions.missingHandle')
+                    ->build(),
+            ];
+        }
+
+        $args = $node->getArgs();
+        if ($methodName === 'runIf' || $methodName === 'runUnless') {
+            $args = array_slice($args, 1);
+        }
```
trevor-cortex: Two edge cases with the positional slice:

1. **Zero-arg `runIf()`/`runUnless()`**: `array_slice([], 1)` is `[]`, then `$reportedArgCount = 0 + 1 = 1`, so `AddAction::runIf()` reports "expects exactly 3 arguments, 1 given" when 0 were given. PHPStan's native check already reports the missing `$boolean` parameter here, so an early `return []` when `count($args) === 0` for the conditional variants would avoid the misleading count (and a duplicate error).
2. **Named condition argument**: `AddAction::runIf(a: 1, boolean: true)`-style calls slice off the *first array element* rather than the condition itself, so the `boolean:` arg leaks into the handle-parameter matching (it just fails the name lookup and is skipped, but the count is off by one). Filtering out the arg named `boolean` — or the first *positional* arg — when slicing would make this robust.

Both are edge cases, so fine to defer, but worth a fixture if you fix them.

src/PHPStan/RunReturnTypeExtension.php
```diff
@@ -0,0 +1,66 @@
...
+        $handleReturnType = $variant->getReturnType();
+
+        if ($methodName === 'run') {
+            return $handleReturnType;
+        }
+
+        return TypeCombinator::union(
+            $handleReturnType,
+            new ObjectType('Illuminate\Support\Fluent'),
+        );
```
trevor-cortex: Nice-to-have: when the condition is a known constant, you could narrow instead of always unioning — `runIf(true, ...)` can return just `handle()`'s type and `runIf(false, ...)` just `Fluent` (inverted for `runUnless`). `$scope->getType($condition)->isTrue()->yes()` / `->isFalse()->yes()` gets you there. Would change a few `assertType` fixtures that currently pass literal `true`/`false`, so maybe one for the follow-up PR.

trevor-cortex (approved):
## Summary

This adds a PHPStan extension that makes `run()`, `runIf()`, and `runUnless()` transparent proxies to `handle()` for static analysis, closing #260:

- `ActionHelper` centralizes resolution: identifies the proxy call, resolves the caller to a single class reflection using the `AsObject` trait, and fetches `handle()`.
- `RunReturnTypeExtension` (via `ExpressionTypeResolverExtension`) propagates `handle()`'s return type, unioning with `Fluent` for the conditional variants. Using the expression-type hook rather than `DynamicStaticMethodReturnTypeExtension` is the right call here since there's no common base class to key on — only a trait.
- `RunParameterValidationRule` re-checks argument count, optionality, variadics, named args, and types against `handle()`'s signature, with counts reported relative to the actual call site (condition included).
- `RunThrowTypeExtension` forwards `handle()`'s `@throws` to callers.

Test coverage is solid: type-inference assertions, rule error assertions with exact messages/lines, and throw propagation via PHPStan's own `MissingCheckedExceptionInFunctionThrowsRule`. Notably, the fixtures use `AsAction` (which pulls in `AsObject` transitively), so the trait-of-trait resolution path in `hasTraitUse` is exercised by the tests — good.

## Things to watch

1. **`runIf`/`runUnless` condition-arg slicing** is positional and assumes the condition is present — see inline comments for two edge cases (zero-arg calls and a named `boolean:` argument).
2. **Fixtures in the autoload classmap**: the fixture files contain top-level executable calls; see inline comment on `composer.json`.
3. **Dynamic class-string callers** (`$action::run()` where `$action` is `class-string<SomeAction>`) fall through silently — see inline comment; easy win if you want it.

## Notes for subsequent reviewers / follow-ups

- The setup instructions (manual `extension.neon` include for users without `phpstan/extension-installer`) live only in the PR description right now — worth adding to the README/docs site alongside this release.
- The rule is registered unconditionally in `extension.neon`, so it runs at every PHPStan level. Type checks go through `RuleLevelHelper`, which respects the user's level-derived strictness parameters, so this should behave consistently with native checks — but it's worth a sanity check in a real consuming project at a low level.
- Scoping to `run()`/`runIf()`/`runUnless()` with job/controller methods deferred to a follow-up is explicitly called out in the description and seems like the right incremental approach.

lorisleiva:
@pindab0ter Looks like it's mainly minor comments from Trevor. Lmk if you wanna tackle any of these otherwise I'm happy to merge as-is. Should we do a minor bump since it's technically a new feature? Or do you see this more as a patch?

src/PHPStan/RunParameterValidationRule.php
pindab0ter: Conditional runs (`runIf`/`runUnless`) are not valid when run with 0 arguments. I've updated the checks to reflect this.

composer.json
pindab0ter: 1. Tested it: dropping the classmap fails 22 of 23 tests in tests/PHPStan (Class …\Fixtures\ThrowingAction not found in ReflectionProvider. Configure "autoload-dev" …). Static reflection covers the analysed file, but gatherAssertTypes() resolves fixture classes through the autoloader-backed ReflectionProvider.

2. It fired rather than staying latent: class_exists('…\Fixtures\AddAction') fatalled with ArgumentCountError. Fixed in fe33a0e by moving the file-scope statements into named functions, matching run-throw-types.php.

pindab0ter:
I've resolved all Trevor's feedback and used the opportunity to give it another thorough pass in which I fixed a number of edge cases.

Minor version sounds great for this PR.

trevor-cortex (approved):
## Re-review

Re-fetched all diffs — this pass addresses everything from my previous review, plus a bit more. Still approved.

### What changed since last time

- **Class-string callers**: `resolveActionClass` now routes non-`Name` callers through `getObjectTypeOrClassStringObjectType()`, and `viaClassStringVariable()` in the return-type fixture covers it. ✅
- **Zero-arg conditional calls**: `findConditionArgIndex` bails on empty args, and the rule now reports full call arity (`runIf() expects exactly 3 arguments, 0 given`) with a fixture comment explaining the choice. The `NoParamsAction::runIf()` case correctly stays silent and defers to PHPStan's native missing-`$boolean` check. ✅
- **Named `boolean:` argument**: condition detection now falls back to a name-based lookup, with `namedConditionArgument()` fixtures verifying both stripping and the `#2 $a` index reporting. ✅
- **Constant-condition narrowing**: `runIf(true, ...)` / `runUnless(false, ...)` now narrow to `handle()`'s return type, and the known-skip case narrows to `Fluent`. The `void → null` handling in the union case is a nice catch — `Fluent|null` is exactly what the runtime produces. ✅
- **Implicit throws** (new): `RunThrowTypeExtension` now takes `%exceptions.implicitThrows%` and falls back to `Throwable` instead of accidentally declaring `run()` throw-free when `handle()` has no `@throws` tag. The dedicated `CatchWithUnthrownExceptionRule`-based test proving the catches aren't dead is the right way to lock this in.

### Additional hardening I hadn't asked for, all verified

- First-class callable guard (`run(...)`) before `getArgs()`.
- Unpacked-argument bail-out in both the rule and the condition-arg detection, with fixtures.
- `hasNativeMethod('handle')` instead of `hasMethod()`, so `@method`-annotated virtual `handle()` is treated as "unknown signature" rather than "missing method" — and the missing-handle error is suppressed for `static::run()` in abstract bases where a subclass supplies `handle()` (covered by `BaseAction`/`ConcreteAction`).
- Named-argument matching against `handle()`'s parameters, including the variadic-collects-any-name case.

### Remaining notes (unchanged from last review, non-blocking)

- The fixture `classmap` entry in `composer.json` is still in place — the fixtures no longer contain top-level executable code, so the footgun I flagged is gone and this is now just a mild redundancy. Fine either way.
- README/docs setup instructions should land alongside the release.
- Job/controller-method coverage deferred to a follow-up remains the right incremental scope.

lorisleiva (approved):
Thank you! I'll publish a new minor today.

Issue #117, opened 2021-05-25
IDE Helpers Prototype


Wulfheart:
### Problem
Given the following action:

```php
class SendTeamReportEmail
{
    use AsAction;

    public function handle(Team $team, bool $fullReport): void
    {
        // ...
    }
}
```

When I'm trying to e.g. dispatch a job via `SendTeamReportEmail::dispatch()` I don't get method signature.

![image](https://user-images.githubusercontent.com/25671390/119563404-c9eef180-bda7-11eb-98a9-578e2d04d346.png)

Therefore I wrote a package which analyzes all actions in a project and creates an ide-helper file like https://github.com/barryvdh/laravel-ide-helper. 

### Installation

```
composer require --dev wulfheart/laravel-actions-ide-helper
```

### Usage
```
php artisan ide-helper:actions
```

### Result

![image](https://user-images.githubusercontent.com/25671390/119564155-c0b25480-bda8-11eb-8bab-e3cf5a58fdd1.png)

### Limitations

It is just a prototype. Therefore the package currently has some limitations.

#### Ignoring the decorator functions

```php
class SendTeamReportEmail
{
    use AsAction;

    public function handle(Team $team, bool $fullReport): void
    {
        // Prepare report and send it to all $team->users.
    }

    public function asJob(Team $team): void
    {
        $this->handle($team, true);
    }
}
```

If I understand the inner workings of Laravel Actions correctly this action has now some other parameters applied if it is dispatched as Job. Currently it assumes the parameters of the `handle` function. I will have to fix this.

#### No default value

If a parameter has default value (e.g. `bool $fullReport = true`)  the default value doesn't get recognized. This will get fixed later.

### Opinion
@lorisleiva What is your opinion on this? Do you think this could be an addition to Laravel Actions or a separate package to improve the developer experience?

Wulfheart:
Update:
The package now supports autocompletion for some decorator functions in an action.

![image](https://user-images.githubusercontent.com/25671390/119736051-78b13180-be7d-11eb-9fe6-d06b2b5d5b04.png)

Furthermore are the different decorators in an action recognized and the parameters of the accessor funtions like `makeJob()` are set properly.
Default types are working now.

lorisleiva:
Hey Alex 👋 

I really like this! I currently fix this for the `run` method manually by adding a `@method` docblock on the class itself that reflects the `handle` method parameters and return type.

Having something that generates it for me for all methods supported by Laravel Actions is really great!

[...]

PR #339, opened 2026-03-29
IDE helper as first party feature

adrum:
I would like to make this a first party feature. [I will maintain it.](https://github.com/lorisleiva/laravel-actions/issues/117#issuecomment-1890907488) It currently is just a straight transfer of [the original repo](https://github.com/Wulfheart/laravel-actions-ide-helper/) with updated namespace. 

I [periodically submit PRs](https://github.com/Wulfheart/laravel-actions-ide-helper/pulls?q=is%3Apr+author%3Aadrum+) to the original repo, but it would be nice to consolidate these two.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,16 @@
     "require": {
         "php": "^8.2",
         "illuminate/contracts": "^11.0|^12.0|^13.0",
-        "lorisleiva/lody": "^0.7"
+        "lorisleiva/lody": "^0.7",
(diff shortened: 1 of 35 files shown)


lorisleiva:
Hi there, thanks for the PR. Tbh, I'm not sure how I feel about adding all that IDE-specific code to this package. I quite like the separation of concerns that a distinct package brings. Happy to re-evaluate if there are pain points I'm missing beside just adding a new dependency.

PR #340, opened 2026-04-28
Support Laravel 13.6 DebounceFor attribute.

huy-tran:
First off, thank you for building and maintaining laravel-actions. We've been using it in almost all of our Laravel projects and it's become an indispensable part of how we structure business logic.

### Summary
Laravel 13.6 introduced the `#[DebounceFor]` PHP attribute for job debouncing. This PR adds support for it in laravel-actions by following the same pattern already established by UniqueJobDecorator.

### Usage

```php
use Illuminate\Queue\Attributes\DebounceFor;
use Lorisleiva\Actions\Concerns\AsJob;

#[DebounceFor(30)]
class SyncUserToMailchimp
{
      use AsJob;

      public string $jobDebounceId; // or getJobDebounceId(): string

      public function handle(User $user): void { ... }
}
```

### Known limitation
`maxWait` from `#[DebounceFor(30, maxWait: 60)]` on the action class is not forwarded. `DebounceLock::getMaxDebounceWait()` reads `maxWait` directly from the job's PHP attribute (not via `getAttributeValue`), so it always sees the decorator's dummy `#[DebounceFor(0)]` with `maxWait = null`. Resolving this would require either a change in Laravel core or a custom `PendingDispatch` subclass.

Thanks again ^^


huy-tran:
Hi @lorisleiva would you mind reviewing this PR. Kind regards,

patrickomeara:
I'm also looking for this feature @lorisleiva 

API looks good, nice work @huy-tran

vpratfr:
Can you update that PR according to the latest branch version so we can get the CI checks going?

huy-tran:
@vpratfr Done 🥳  Rebased onto the latest main (was 11 commits behind, which is why the new PHPStan CI workflow wasn't showing up) and force-pushed. Also fixed a PHPStan error the new level 5 check would've caught in the new tests (a property access on the base **JobDecorator** type instead of **DebounceJobDecorator**).

PR #341, opened 2026-05-19
fix: handle missing Actions folder gracefully

### LINKED ISSUE #291: You must call one of in() or append() methods before iterating over a Finder.
In the latest versions of Laravel, if you install the Actions package and register in your app service provider boot method `Actions::registerCommands();`, it will return a LogicException: "You must call one of in() or append() methods before iterating over a Finder." IF you do not have any actions created yet and you try to run any artist commands, composer installs, etc. It tries to iterate on the Actions folder but it doesn't not exist yet. 

I will say that i am on windows, so that may behave differently than linux/mac for this scenario. But as soon as i have the actions folder i can then do the registerCommands() and it will work fine when i try to run any artisan command

YoussefMansour9:
When Actions::registerCommands() or Actions::registerRoutes() is called before the Actions folder exists, Lody throws a LogicException because the Symfony Finder is iterated without any directories set.

This fix adds a check to ensure the path exists before attempting to discover classes, allowing the methods to be called safely even when the folder doesn't exist yet.

Fixes #291

### DIFF
--- a/src/ActionManager.php
+++ b/src/ActionManager.php
@@ -158,23 +158,39 @@ class ActionManager
 
     public function registerRoutes(array | string $paths = 'app/Actions'): void
     {
-        Lody::classes($paths)
-            ->isNotAbstract()
-            ->hasTrait(AsController::class)
-            ->hasStaticMethod('routes')
-            ->each(fn (string $classname) => $this->registerRoutesForAction($classname));
+        $paths = is_array($paths) ? $paths : [$paths];
+
+        foreach ($paths as $path) {
+            if (! is_dir(base_path($path))) {
+                continue;
+            }
+
+            Lody::classes($path)
+                ->isNotAbstract()
+                ->hasTrait(AsController::class)
+                ->hasStaticMethod('routes')
+                ->each(fn (string $classname) => $this->registerRoutesForAction($classname));
+        }
     }
 
     public function registerCommands(array | string $paths = 'app/Actions'): void
     {
-        Lody::classes($paths)
-            ->isNotAbstract()
-            ->hasTrait(AsCommand::class)
-            ->filter(function (string $classname): bool {
-                return property_exists($classname, 'commandSignature')
-                    || method_exists($classname, 'getCommandSignature');
-            })
-            ->each(fn (string $classname) => $this->registerCommandsForAction($classname));
+        $paths = is_array($paths) ? $paths : [$paths];
+
+        foreach ($paths as $path) {
+            if (! is_dir(base_path($path))) {
+                continue;
+            }
+
+            Lody::classes($path)
+                ->isNotAbstract()
+                ->hasTrait(AsCommand::class)
+                ->filter(function (string $classname): bool {
+                    return property_exists($classname, 'commandSignature')
+                        || method_exists($classname, 'getCommandSignature');
+                })
+                ->each(fn (string $classname) => $this->registerCommandsForAction($classname));
+        }
     }
 
     public function registerRoutesForAction(string $className): void


vpratfr (changes requested):
Thanks for tackling this @YoussefMansour9 — the bug is real and worth fixing, but I don't think the patch can go in as-is: it trades the crash for a silent failure on absolute paths. Could you address the points below?

## Root cause

Worth restating, because it points at a cleaner fix. `LodyManager::resolvePaths()` already drops non-existent directories:

```php
return Collection::wrap($paths)
    ->map(fn (string $path) => $this->resolvePath($path))
    ->unique()
    ->filter(fn (string $path) => is_dir($path))   // <- already guards
    ->values()->toArray();
```

So the `LogicException` isn't "the directory is missing" — it's that the *filtered list becomes empty*, and `Finder::create()->in([])` throws on iteration. That also explains why the same crash reaches consumers that never touch `ActionManager` (see the ide-helper stack trace in #291).

## Blocking

### 1. `base_path($path)` breaks absolute paths

Lody's `resolvePath()` deliberately passes absolute paths and Windows drive letters straight through:

```php
if ($startsWithDirectorySeparator || $startsWithWindowsDisk) {
    return $path;
}
return $this->getBasePath($path);
```

This PR bypasses that and always prepends the base path. On Laravel 11+ (which is the floor here: `illuminate/contracts: ^11.0|^12.0|^13.0`), `base_path()` goes through `join_paths()`, which `ltrim`s the leading separator and concatenates:

```
base_path('/var/www/app/Actions')  ->  /home/user/project/var/www/app/Actions
```

`is_dir()` on that is `false`, so `continue` fires and **the path is silently skipped**. Anyone calling `Actions::registerRoutes(app_path('Actions'))` or `registerCommands(__DIR__.'/Actions')` loses all route/command registration with no error at all — strictly worse than the exception it replaces. Same breakage for `C:\...`, which is unfortunate given the reporter in #291 was on Windows.

### 2. Bypasses Lody's configurable path resolution

`Lody::resolvePathUsing()` and `Lody::setBasePath()` are public API. Hardcoding `base_path()` means the existence check and the actual discovery can disagree about what a given path means.

## Non-blocking

### 3. Loses `unique()` and cross-path ordering

Replacing one `Lody::classes()` call over all paths with one call *per path* means `resolvePaths()`'s `->unique()` no longer applies, so a duplicated path entry now registers routes twice (`$className::routes()` invoked twice). `sortByName()` also becomes per-directory instead of global, which changes registration order when several paths are passed — and Laravel routing is first-match-wins.

### 4. No tests

This is trivially testable with the existing Pest + Testbench setup: under Testbench, `base_path('app/Actions')` doesn't exist, so `main` throws today on `Actions::registerCommands()` with default arguments.

```php
it('does not throw when the actions folder does not exist', function () {
    Actions::registerCommands('app/DoesNotExist');
    Actions::registerRoutes('app/DoesNotExist');
})->throwsNoExceptions();
```

Please also add one asserting that an existing fixture directory still discovers its actions — that's the test that would have caught point 1.

### 5. Duplicated guard

The same check is copy-pasted into both methods; worth extracting.

## Suggested shape

Reuse Lody's own resolver, keep a single `Lody::classes()` call, and short-circuit on empty:

```php
use Illuminate\Support\Arr;
use Lorisleiva\Lody\ClassnameLazyCollection;
use Lorisleiva\Lody\LodyManager;

public function registerRoutes(array | string $paths = 'app/Actions'): void
{
    $this->discoverActions($paths)
        ->hasTrait(AsController::class)
        ->hasStaticMethod('routes')
        ->each(fn (string $classname) => $this->registerRoutesForAction($classname));
}

public function registerCommands(array | string $paths = 'app/Actions'): void
{
    $this->discoverActions($paths)
        ->hasTrait(AsCommand::class)
        ->filter(function (string $classname): bool {
            return property_exists($classname, 'commandSignature')
                || method_exists($classname, 'getCommandSignature');
        })
        ->each(fn (string $classname) => $this->registerCommandsForAction($classname));
}

protected function discoverActions(array | string $paths): ClassnameLazyCollection
{
    $lody = app(LodyManager::class);

    $existingPaths = collect(Arr::wrap($paths))
        ->filter(fn (string $path) => is_dir($lody->resolvePath($path)))
        ->values()
        ->all();

    return $existingPaths === []
        ? ClassnameLazyCollection::make([])
        : Lody::classes($existingPaths)->isNotAbstract();
}
```

Resolve through `app(LodyManager::class)` rather than `Lody::resolvePath()` — `resolvePath` isn't declared in the facade's `@method` docblock and Larastan runs in CI.

## Note for @lorisleiva

The equivalent one-liner in `lorisleiva/lody` would cover every consumer rather than just this package:

```php
public function files(array | string $paths, bool $recursive = true, bool $hidden = false): FileLazyCollection
{
    $resolved = $this->resolvePaths($paths);

    if ($resolved === []) {
        return FileLazyCollection::make([]);
    }
    // ...
}
```

That's what actually fixes the ide-helper crash reported in #291. A defensive guard here is still worth having, since it works against the already-released Lody `^0.7` constraint — but both would be ideal.

One last thing: CI reports no checks run on this branch. Once the above is addressed, it'd be good to see the suite green.

vpratfr:
Also, can you update that PR according to the latest branch version so we can get the CI checks going?

lorisleiva:
If the fix is more elegant in Lody, let's do it there. 👌

vpratfr:
> If the fix is more elegant in Lody, let's do it there. 👌

Ok. Let me know when done so I can it here.

lorisleiva:
Sorry I won't have bandwidth for this for a while but happy to accept a PR there and publish a patch bump.

vpratfr:
> Sorry I won't have bandwidth for this for a while but happy to accept a PR there and publish a patch bump.

Submitted: https://github.com/lorisleiva/lody/pull/13 @lorisleiva

vpratfr:
https://github.com/lorisleiva/lody/pull/13 fixes the issue and has been merged. No need for that one then.

PR #346, opened 2026-06-30
Fix AsListener not recognized when event implements ShouldDispatchAfterCommit

### LINKED ISSUE #345: AsListener breaks when event implements ShouldDispatchAfterCommit inside a database transaction
## Description

When an event implements `ShouldDispatchAfterCommit` and is dispatched inside an active database transaction, actions registered as listeners are called incorrectly: the `ListenerDecorator` is not applied, and Laravel calls `handle()` directly with the raw event object instead of using the action's `asListener()` method (or resolving typed dependencies via `handle()`).

## Root Cause

`ListenerDesignPattern::recognizeFrame()` detects the listener context by scanning the call stack for `Dispatcher::dispatch`. This works for normal event dispatching.

However, in Laravel 12, when an event implements `ShouldDispatchAfterCommit` and there is an active transaction, `Dispatcher::dispatch()` defers execution via a callback registered with `DatabaseTransactionsManager`:

```php
$transactions->addCallback(
    fn () => $this->invokeListeners($parsedEvent, $parsedPayload, $halt)
);
```

`dispatch()` returns immediately, unwinding the stack. When the transaction commits and the callback fires, the call stack no longer contains `Dispatcher::dispatch` — only `Dispatcher::invokeListeners` and the transactions manager frames. As a result, `recognizeFrame()` finds no match, no decorator is applied, and Laravel calls `handle()` directly with the event object.

## Reproduction

```php
class UploadStatusChanged implements ShouldDispatchAfterCommit
{
    public function __construct(public Upload $upload) {}
}

class HandleUploadStatusChanged
{
    use AsListener, AsObject;

    public function handle(Upload $upload): void
    {
        // process upload...
    }

    public function asListener(UploadStatusChanged $event): void
    {
        $this->handle($event->upload);
    }
}

// Registered as:
Event::listen(UploadStatusChanged::class, HandleUploadStatusChanged::class);

// When dispatched inside a transaction:
DB::beginTransaction();
Event::dispatch(new UploadStatusChanged($upload));
DB::commit(); // ← handle() receives the UploadStatusChanged event object instead of asListener() being called
```

## Fix

Add `Dispatcher::invokeListeners` to the recognized frames in `ListenerDesignPattern::recognizeFrame()`. This method is the convergence point for both the normal and `ShouldDispatchAfterCommit` dispatch paths in Laravel 12:

```php
public function recognizeFrame(BacktraceFrame $frame): bool
{
    return $frame->matches(Dispatcher::class, 'dispatch')
        || $frame->matches(Dispatcher::class, 'invokeListeners') // fix
        || $frame->matches(Dispatcher::class, 'handlerWantsToBeQueued')
        || $frame->matches(CallQueuedListener::class, 'handle')
        || $frame->matches(CallQueuedListener::class, 'failed');
}
```

`invokeListeners` is a protected method introduced alongside the `ShouldDispatchAfterCommit` mechanism. It does not appear in controller or command backtraces, so the change has no impact on other design patterns.

## Environment

- PHP 8.4
- Laravel 12
- lorisleiva/laravel-actions (latest)

vpratfr:
## Problem

When an event implements `ShouldDispatchAfterCommit` and is dispatched inside an active database transaction, the `ListenerDecorator` is never applied. Laravel ends up calling `handle()` directly with the raw event object instead of going through `asListener()` or resolving typed dependencies.

**Minimal reproduction:**

```php
class UploadStatusChanged implements ShouldDispatchAfterCommit
{
    public function __construct(public Upload $upload) {}
}

class HandleUploadStatusChanged
{
    use AsListener, AsObject;

    public function handle(Upload $upload): void { /* ... */ }

    public function asListener(UploadStatusChanged $event): void
    {
        $this->handle($event->upload);
    }
}

Event::listen(UploadStatusChanged::class, HandleUploadStatusChanged::class);

DB::beginTransaction();
Event::dispatch(new UploadStatusChanged($upload));
DB::commit(); // ← asListener() is never called; handle() receives the event object
```

## Root Cause

`ListenerDesignPattern::recognizeFrame()` detects the listener context by scanning the call stack for `Dispatcher::dispatch`. This works for normal dispatches.

However, for `ShouldDispatchAfterCommit` events inside a transaction, Laravel 12 defers execution via a stored callback:

```php
// Dispatcher::dispatch()
$transactions->addCallback(
    fn () => $this->invokeListeners($parsedEvent, $parsedPayload, $halt)
);
return null; // dispatch() exits here, stack is unwound
```

When the transaction commits and the callback fires, `Dispatcher::dispatch` is no longer on the call stack. `recognizeFrame()` finds no match, no decorator is applied, and the action falls back to being called as a plain object.

## Fix

Add `Dispatcher::invokeListeners` to the recognized frames. It is the convergence point for both dispatch paths in Laravel 12:

```php
// Normal path
return $this->invokeListeners($parsedEvent, $parsedPayload, $halt);

// ShouldDispatchAfterCommit path
$transactions->addCallback(fn () => $this->invokeListeners(...));
```

`invokeListeners` is a protected method specific to event dispatching. It does not appear in controller (`Route::getController`) or command (`Application::resolve`) call stacks, so the change does not affect other design patterns.

## Changes

- `src/DesignPatterns/ListenerDesignPattern.php` — add `Dispatcher::invokeListeners` to `recognizeFrame()`
- `tests/AsListenerWithShouldDispatchAfterCommitTest.php` — new test covering object use, dispatch outside a transaction, and dispatch inside a committed transaction
- `tests/Stubs/OperationRequestedAfterCommitEvent.php` — event stub implementing `ShouldDispatchAfterCommit`

Closes #345

### DIFF
--- a/src/DesignPatterns/ListenerDesignPattern.php
+++ b/src/DesignPatterns/ListenerDesignPattern.php
@@ -18,6 +18,7 @@ class ListenerDesignPattern extends DesignPattern
     public function recognizeFrame(BacktraceFrame $frame): bool
     {
         return $frame->matches(Dispatcher::class, 'dispatch')
+            || $frame->matches(Dispatcher::class, 'invokeListeners')
             || $frame->matches(Dispatcher::class, 'handlerWantsToBeQueued')
             || $frame->matches(CallQueuedListener::class, 'handle')
             || $frame->matches(CallQueuedListener::class, 'failed');
--- a/tests/AsListenerWithShouldDispatchAfterCommitTest.php
+++ b/tests/AsListenerWithShouldDispatchAfterCommitTest.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Lorisleiva\Actions\Tests;
+
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Event;
+use Lorisleiva\Actions\Concerns\AsListener;
+use Lorisleiva\Actions\Concerns\AsObject;
+use Lorisleiva\Actions\Tests\Stubs\OperationRequestedAfterCommitEvent;
+
+class AsListenerWithShouldDispatchAfterCommitTest
+{
+    use AsObject;
+    use AsListener;
+
+    public static ?int $latestResult;
+
+    public function handle(int $left, int $right, bool $addition = true): int
+    {
+        return $addition ? $left + $right : $left - $right;
+    }
+
+    public function asListener(OperationRequestedAfterCommitEvent $event): void
+    {
+        static::$latestResult = $this->handle(
+            $event->left,
+            $event->right,
+            $event->operation === 'addition'
+        );
+    }
+}
+
+beforeEach(function () {
+    // Given we are listening for the OperationRequestedAfterCommitEvent.
+    Event::listen(OperationRequestedAfterCommitEvent::class, AsListenerWithShouldDispatchAfterCommitTest::class);
+
+    // And reset the latest result between tests.
+    AsListenerWithShouldDispatchAfterCommitTest::$latestResult = null;
+});
+
+it('works as an object even when the event implements ShouldDispatchAfterCommit', function () {
+    // When we run the action as an object.
+    $result = AsListenerWithShouldDispatchAfterCommitTest::run(5, 3, false);
+
+    // Then we get the expected result.
+    expect($result)->toBe(2);
+
+    // And the `asListener` method was not called.
+    expect(AsListenerWithShouldDispatchAfterCommitTest::$latestResult)->toBeNull();
+});
+
+it('calls asListener() when event implements ShouldDispatchAfterCommit and is dispatched outside a transaction', function () {
+    // When we dispatch the event outside of a transaction.
+    Event::dispatch(new OperationRequestedAfterCommitEvent('addition', 1, 2));
+
+    // Then asListener() was called via the ListenerDecorator.
+    expect(AsListenerWithShouldDispatchAfterCommitTest::$latestResult)->toBe(3);
+});
+
+it('calls asListener() when event implements ShouldDispatchAfterCommit and is dispatched inside a transaction', function () {
+    // When we dispatch the event inside a transaction and commit.
+    DB::beginTransaction();
+    Event::dispatch(new OperationRequestedAfterCommitEvent('addition', 1, 2));
+
+    // Then asListener() has not been called yet (deferred until commit).
+    expect(AsListenerWithShouldDispatchAfterCommitTest::$latestResult)->toBeNull();
+
+    DB::commit();
+
+    // Then asListener() was called via the ListenerDecorator after the commit.
+    expect(AsListenerWithShouldDispatchAfterCommitTest::$latestResult)->toBe(3);
+});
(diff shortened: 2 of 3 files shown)


vpratfr:
Hi @lorisleiva 

Currently that is a blocking issue in a large project of us.

Any idea if that could be merged along with some other nice PRs that are currently queued? (PHPStan support, etc.)? My fork is currently in production on that project and running fine. 

CI should turn green, and if anything is missing I'd be happy to fix it.

If you are looking for maintainers to review some PRs I could help a little bit.

Regards,

Vincent

trevor-cortex (approved):
## Summary

Fixes a real bug where actions using `AsListener` bypass the `ListenerDecorator` when the event implements `ShouldDispatchAfterCommit` and is dispatched inside an active transaction. `handle()` ends up being called with the raw event instead of `asListener()` being invoked with typed dependencies.

Root cause is accurate: `ListenerDesignPattern::recognizeFrame()` scans the call stack for `Dispatcher::dispatch`, but when Laravel defers execution via `$transactions->addCallback(...)`, `dispatch` has already unwound by the time listeners are invoked at commit time.

The fix — adding `Dispatcher::invokeListeners` to the recognized frames — is well-chosen:

- `invokeListeners` is the convergence point for both the normal path (`return $this->invokeListeners(...)` at the tail of `dispatch`) and the deferred path (`fn () => $this->invokeListeners(...)` stored on the transaction manager).
- It's a `protected` method specific to `Illuminate\Events\Dispatcher`, so it won't collide with the controller (`Route::getController`) or command (`Application::resolve`) frames that other design patterns rely on.
- Minimal, one-line change; low blast radius.

## Tests

Good coverage of the three relevant scenarios:
1. Action still works as a plain object when the event implements `ShouldDispatchAfterCommit` (regression guard for the `run()` path).
2. `asListener()` is called for dispatches outside a transaction (already worked, but nice to lock in).
3. `asListener()` is deferred until commit for dispatches inside a transaction (the actual bug fix).

Style matches the existing `AsListenerTest.php` — Pest with a class + trait, `beforeEach` for setup, static property for capturing results. The new `OperationRequestedAfterCommitEvent` stub is a minimal parallel to `OperationRequestedEvent` with the marker interface added, which reads clearly.

The transaction test relies on Testbench's default in-memory sqlite being wired up (via `database.default = sqlite` in `TestCase`), and since the event doesn't touch any tables, only the transaction manager state matters — that should be robust.

## Notes for subsequent reviewers

- Worth confirming behavior on Laravel 11 as well, since the package supports `illuminate/contracts` `^11.0|^12.0|^13.0`. The `ShouldDispatchAfterCommit` + `invokeListeners` shape has been stable across these versions, but a quick CI run across the matrix is the definitive check.
- Queued listener path (`ShouldQueue` + `ShouldDispatchAfterCommit`) is not exercised by the new tests, but it goes through `CallQueuedListener::handle` which was already recognized, so it should be unaffected.
- No rollback test — i.e. that `asListener()` is *not* called if the transaction is rolled back. That's really Laravel's responsibility (it's what `addCallback` guarantees), so it's fine to leave out, but could be added as an extra safety net if desired.

LGTM.

lorisleiva (approved):
Hi there, sorry for the late review! Thank you for fixing this. I'll release a new patch version asap.

PR #347, opened 2026-08-17
Add root PHPStan configuration at level 5

pindab0ter:
Adds a root `phpstan.neon.dist` at level 5 so the package analyses itself, including the extension from #338. Adds `larastan/larastan`, `phpstan/phpstan-phpunit` and `phpstan/extension-installer`, and runs PHPStan in CI.

Builds on #338, so the diff includes its commits until that merges.

Level 5 rather than higher because level 6 takes it from 15 to 187 errors, nearly all `missingType.*` on unannotated signatures.

Getting to a clean run needed some type declaration fixes across `src` and three test files. No runtime behaviour changes. The one worth calling out: `ActionManager::$uniqueJobDecorator` was annotated `class-string<JobDecorator&ShouldBeUnique>` while `makeUniqueJob()` returns `UniqueJobDecorator`, so a custom decorator implementing `ShouldBeUnique` without extending `UniqueJobDecorator` would have violated the declared return type.

Also declares `friendsofphp/php-cs-fixer` in `require-dev`. It was installed and present in the lock but never declared, so `composer validate` flagged the lock as stale and `composer update` dropped it.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -19,9 +19,9 @@ jobs:
           extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, intl
           coverage: none
 
-      # Analysis runs on a single dependency set. The baseline records line
-      # numbers and inferred types, so a prefer-lowest resolution would report
-      # errors that do not match it.
+      # Analysis runs on a single dependency set. Larastan infers different
+      # types across Laravel versions, so a prefer-lowest resolution would
+      # report errors the configuration does not account for.
       - name: Install dependencies
         run: composer update --prefer-dist --no-interaction
 
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,29 @@
+name: PHPStan
+
+on: [push, pull_request]
+
+jobs:
+  phpstan:
+    runs-on: ubuntu-latest
+
+    name: PHPStan
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "8.4"
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, intl
+          coverage: none
+
+      # Analysis runs on a single dependency set. The baseline records line
+      # numbers and inferred types, so a prefer-lowest resolution would report
+      # errors that do not match it.
+      - name: Install dependencies
+        run: composer update --prefer-dist --no-interaction
+
+      - name: Run PHPStan
+        run: vendor/bin/phpstan analyse --no-progress
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ build
 composer.lock
 coverage
 docs
+phpstan.neon
 phpunit.xml
 psalm.xml
 vendor
--- a/composer.json
+++ b/composer.json
@@ -27,9 +27,13 @@
         "lorisleiva/lody": "^0.7"
     },
     "require-dev": {
+        "friendsofphp/php-cs-fixer": "^3.0",
+        "larastan/larastan": "^3.0",
         "orchestra/testbench": "^9.0|^10.0|^11.0",
         "pestphp/pest": "^3.0|^4.0",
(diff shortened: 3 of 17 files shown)


lorisleiva:
Oh I've just noticed this follow-up PR sorry. Already published `v2.11.0`. Would you mind rebasing your changes now that #338 is published?

pindab0ter:
Done! And no worries, these changes should not affect any end user as they are only code quality related, so no need to create a new version for this IMO.

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "8.4"
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, intl
+          coverage: none
+
+      # Analysis runs on a single dependency set. The baseline records line
+      # numbers and inferred types, so a prefer-lowest resolution would report
+      # errors that do not match it.
```
trevor-cortex: This comment justifies the single dependency set by referencing a baseline, but the PR ships no `phpstan-baseline.neon` — errors are handled via `ignoreErrors` identifiers/patterns in the neon config, which are line-number-independent. The single-dependency-set choice is still sound (larastan's inferred types shift across Laravel versions), but consider rewording so the comment doesn't reference a baseline that doesn't exist — it'll confuse the next person who greps for one.

trevor-cortex (approved):
Adds a root `phpstan.neon.dist` at level 5 covering `src` and `tests`, wires up larastan/phpstan-phpunit via `extension-installer`, includes the package's own `extension.neon`, and runs the analysis in CI. The type fixes needed for a clean run are all annotation-level or trivially equivalent rewrites — I verified none of them change runtime behaviour:

- `ValidateActions::failedValidation()` / `prepareForValidation()`: return values were already discarded by the only caller (`validate()`), so dropping the `return` is safe.
- `AsJob::assertPushed()`: `Queue::getFacadeRoot()` + `pushed()` is behaviourally identical to the facade call; the `@var QueueFake` assumption matches the pre-existing requirement that the queue be faked.
- `ActionManager::identifyFromBacktrace()`: renaming the loop variable preserves the by-ref out-parameter assignment.
- The `$uniqueJobDecorator` narrowing to `class-string<UniqueJobDecorator>` is the right call — `makeUniqueJob()` already declared `UniqueJobDecorator` as its return type, so the old `JobDecorator&ShouldBeUnique` annotation permitted classes that would have violated it. Worth noting it does narrow the documented contract for `useUniqueJobDecorator()`: a custom decorator implementing `ShouldBeUnique` without extending `UniqueJobDecorator` was previously blessed by the docblock (even if broken in practice). If the published docs describe the interface-based approach, they may want a matching touch-up.

The neon config is well-commented — each `ignoreErrors` entry explains *why* the error is a false positive rather than just suppressing it, and the `method.notFound` ignore for `handle()` in the concerns correctly avoids the `@method` tag that would break the `laravelActions.missingHandle` rule. Excluding `tests/PHPStan/Fixtures` from analysis (but not scanning) is the right choice for deliberately-invalid fixture code.

Notes for subsequent reviewers:

- The CI job deliberately analyses a single dependency set (latest stable on PHP 8.4) rather than the test matrix — reasonable, since larastan output varies across Laravel versions. One inline note about the comment justifying this.
- `phpstan.neon` in `.gitignore` follows the standard `.dist` override pattern, consistent with `phpunit.xml`.
- The `(int)` casts in the two command tests change `null` to `0` if an argument were missing, but both tests always pass the arguments, so this is inert.

lorisleiva (approved):
Thanks! @pindab0ter Lmk if you wanna tackle some of Trevor's nits otherwise I'm happy to merge as-is.

pindab0ter:
I resolved Trevor's only nit 👍

vpratfr:
I'll go ahead and merge this PR if that's OK for you @lorisleiva : all checks pass, author has resolved all comments.

This should help with the other pending PRs: that will add a PHPStan check to all PRs.

vpratfr:
I will not create the minor release yet, will wait for a real new bugfix to create one, giving some time to revert this PR if @lorisleiva wants to.

lorisleiva:
Thanks for the merge @vpratfr, I don't think we need to release anything for this PR though as @pindab0ter mentioned since these are just internal changes that don't affect consumers.

